PR #5, opened 2017-05-31
Replicate all main dataset models

govuklaurence:
These are all the dataset models from the django app, plus one statistics model. The models are there but they are not linked up properly yet i.e. no belongs_to or has_many


hannako:
@govuklaurence As discussed I will do this on my branch.

PR #6, opened 2017-05-31
Generate devise models

hannako:
- Adds Devise to gemfile
- Generates the devise PublishingUser model.
- NB: I wasn't sure how to implement the following python:

```PublishingUser._meta.get_field('email')._unique = True
PublishingUser.REQUIRED_FIELDS.remove("email")
PublishingUser.REQUIRED_FIELDS.append("username")
```

which is at the bottom of the original model.


app/models/publishing_user.rb
```diff
@@ -0,0 +1,15 @@
...
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable and :omniauthable
+  devise :database_authenticatable, :registerable,
+         :recoverable, :rememberable, :trackable, :validatable
+
+  def self.primary_organisation
+    organisations.order_by('name').first
+  end
+
+  def self.__repr__
```
govuklaurence: There's no `__repr__` in ruby, you can use `to_s` instead -- but in this case i think it's fine just to return the model as it currently is

govuklaurence: (i.e. this method is not needed)

app/models/publishing_user.rb
```diff
@@ -0,0 +1,15 @@
+class PublishingUser < ApplicationRecord
+  # Include default devise modules. Others available are:
+  # :confirmable, :lockable, :timeoutable and :omniauthable
+  devise :database_authenticatable, :registerable,
+         :recoverable, :rememberable, :trackable, :validatable
+
+  def self.primary_organisation
```
govuklaurence: `self` is implicit in the method declarations here, so you can remove it

app/models/publishing_user.rb
```diff
@@ -0,0 +1,15 @@
...
+    organisations.order_by('name').first
```
govuklaurence: why ordered by name? could we instead include a field on the user for the preferred org?

app/models/publishing_user.rb
```diff
@@ -0,0 +1,15 @@
+class PublishingUser < ApplicationRecord
```
govuklaurence: since there's a many-to-many relationship with organisations, there needs to be a `has_and_belongs_to_many :organisations`

I'll make sure to add the same relations on the org model too

db/migrate/20170531153836_devise_create_publishing_users.rb
```diff
@@ -0,0 +1,44 @@
...
+      t.datetime :reset_password_sent_at
+
+      ## Rememberable
+      t.datetime :remember_created_at
+
+      ## Trackable
+      t.integer  :sign_in_count, default: 0, null: false
+      t.datetime :current_sign_in_at
+      t.datetime :last_sign_in_at
+      t.inet     :current_sign_in_ip
```
govuklaurence: Cool, I didn't know that type existed

govuklaurence:
Don't worry about including

```
PublishingUser.REQUIRED_FIELDS.remove("email")
PublishingUser.REQUIRED_FIELDS.append("username")
```

for now... we can work that in to validations if we need to later

db/migrate/20170601085954_add_primary_organisation_to_publishing_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddPrimaryOrganisationToPublishingUser < ActiveRecord::Migration[5.1]
+  def change
+    add_column :publishing_users, :primary_organisation, :string
```
govuklaurence: let's use a references/id instead here

PR #11, opened 2017-06-01
Front end tests

maxf:
add the front-end tests from publish_data_alpha, and adapt them to Rails.
Missing: adding fixtures to populate the test database with datasets, organisations and users

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -6,12 +6,12 @@ default: &default
 development:
   <<: *default
   database: publish_data_beta_development
-+  url: <%= ENV['DATABASE_URL'] %>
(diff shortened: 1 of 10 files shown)


config/database.yml
```diff
@@ -6,12 +6,12 @@ default: &default
 development:
   <<: *default
   database: publish_data_beta_development
-+  url: <%= ENV['DATABASE_URL'] %>
+  url: <%= ENV['DATABASE_URL'] %>
```
govuklaurence: Just need to remove these

govuklaurence: My bad -- just looked confusing. Approving.

PR #12, opened 2017-06-02
Generate controllers and add routes

govuklaurence:
Adds datasets, tasks, and users as CRUD controllers, and manage as a custom one with routes for manage_own and manage_org


govuklaurence:
The users one does not override the devise controllers -- it should be removed if it is not needed, but I added them in for now just so we can keep the same pattern as django

PR #17, opened 2017-06-05
First pass at getting styles implemented

rossjones:
Uses the govuk_frontend_toolkit_gem and govuk_template to change how
pages are displayed.

Currently broken CSS includes (but probably not limited to):

* Buttons don't seem to pick up colour and styling.
* .masthead in / is not full-width and it should be.

We also have a little hacky css in application.css to hide the govuk
header bar and allow us to display the DGU one.

Will add issues for these so they can be looked at later (unless someone can work out the diff between this and https://github.com/datagovuk/publish_data_alpha which is complicated by the jinja/erb difference).


app/assets/stylesheets/application.css
```diff
@@ -10,6 +10,10 @@
  * files in this directory. Styles in this file should be added after the last require_* statement.
  * It is generally better to create a new file per style scope.
  *
- *= require_tree .
```
govuklaurence: Why was this removed?

app/controllers/home_controller.rb
```diff
@@ -0,0 +1,4 @@
+class HomeController < ApplicationController
```
govuklaurence: On the current implementation the tasks page is the default index page, we could use that controller instead

config/routes.rb
```diff
@@ -1,4 +1,6 @@
 Rails.application.routes.draw do
+  get 'home/index'
```
govuklaurence: May need a `to:` (with root i think it's not required)

config/routes.rb
```diff
@@ -12,6 +14,7 @@
   resources :users
   resources :datasets
 
+  get '', to: 'home#index'
```
govuklaurence: Use

`root to: 'home#index'`

app/assets/stylesheets/application.css
hannako: @rossjones If we don't require the tree how will all of the other stylesheets and assets be compiled?

app/controllers/home_controller.rb
govuklaurence: No worries - brainfart.

app/assets/stylesheets/application.css
rossjones: Don't want the active admin styles to be includes on pages that don't need it.

rossjones: We only require the govuk ones for most pages.

hannako: Unless I misunderstood, Max said he'd customised the css so I should copy it over.

rossjones: Yeah, I added Max's stuff to govuk.scss.erb (so that we can customise the asset paths).

rossjones: Not having ```require_tree .``` does break the CSS for active-admin.  I'll put it back and add a ticket for splitting it out later (possibly).

rossjones: Have moved the generated active_admin.scss to vendor/assets/stylesheets now so it is only picked up when you visit /admin.

govuklaurence: cool... we can go for explicit requires where necessary

app/assets/stylesheets/application.css
```diff
@@ -10,6 +10,10 @@
...
+ *= require govuk
  *= require_self
  */
+
+/* Hide the header from govuk template, make sure the DGU one is shown */
+#global-header { display: none; }
+.global-header-dgu { display: block !important; }
```
govuklaurence: there's only one on the page, we should use an ID

PR #27, opened 2017-06-05
Task model controller views

hannako:
- adds some seed data
- creates task model
- creates basic tasks controller. Helper methods eg `get_tasks_for_user` and groups eg `@datasetsBroken` not yet implemented, just pulling in all datasets/tasks
- creates task views. Filters not yet implemented, just hard coded
- imports show-hide js to a component


db/migrate/20170605140131_create_tasks.rb
```diff
@@ -0,0 +1,15 @@
+class CreateTasks < ActiveRecord::Migration[5.1]
+  def change
+    create_table :tasks do |t|
+      t.references :organisation, foreign_key: true
+      t.integer :quantity, default: 0
+      t.string :required_permission_name
+      t.string :description, limit: 128, default: "", null: true
+      t.string :category, limit: 20
+  #TODO rename to short_name?
```
govuklaurence: Absolutely do

app/controllers/tasks_controller.rb
```diff
@@ -1,4 +1,33 @@
...
+    @organisation = current_publishing_user.primary_organisation
+    @datasets = Dataset.all
+    @tasks = get_tasks_for_organisation(@organisation.name)
+    @datasetsUpdate = @datasets.all
+    @datasetsBroken = @datasets.all
+  end
+
+private
+
+  def get_tasks_for_user(user)
```
govuklaurence: Just fix these two CC issues then we can merge

config/routes.rb
```diff
@@ -7,11 +7,11 @@
       sessions: 'publishing_users/sessions'
     }
 
-  root to: 'tasks#index'
+  root to: 'tasks#my'
```
govuklaurence: This may not merge happily, suggest that you're rebased to the latest master

spec/models/task_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe Task, type: :model do
```
govuklaurence: Just get rid of this file - if it's not doing anything we don't need it. Integration tests are more useful at this stage.

config/routes.rb
rossjones: Maybe we should route tasks to /tasks, we can redirect there from home if user is logged in.

govuklaurence: There's routes for that below, and agree, on home controller we should redirect as a before action

PR #38, opened 2017-06-06
Fix the login template

rossjones:
Fixes the login template to use the govuk styles

### DIFF
--- a/app/views/publishing_users/sessions/new.html.erb
+++ b/app/views/publishing_users/sessions/new.html.erb
@@ -8,7 +8,7 @@
 
   <%= form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
     <div class="form-group">
-      <%= f.label :email %><br />
+      <%= f.label :email_address %><br />
       <%= f.email_field :email, autofocus: true, class: "form-control" %>
     </div>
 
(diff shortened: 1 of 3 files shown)


app/views/publishing_users/sessions/new.html.erb
```diff
@@ -1,26 +1,34 @@
...
+    </div>
   </div>
 
-  <div class="field">
-    <%= f.label :password %><br />
-    <%= f.password_field :password, autocomplete: "off" %>
-  </div>
+  <%= form_for(resource, as: resource_name, url: session_path(resource_name)) do |f| %>
+    <div class="form-group">
+      <%= f.label :email %><br />
```
hannako: on alpha we have email address not email

app/views/publishing_users/sessions/new.html.erb
```diff
@@ -1,26 +1,34 @@
...
+        <%= f.label :remember_me %>
+      </div>
+    <% end -%>
+
+    <div class="actions">
+      <%= f.submit "Sign in", class: "button" %>
+    </div>
+  <% end %>
+
+  <%= render "publishing_users/shared/links" %>
```
hannako: this is pulling in the sign up link, which we don't have on alpha as the user has the create an account option on the landing page. I think maybe we should just delete:

```
<%- if devise_mapping.registerable? && controller_name != 'registrations' %>
  <%= link_to "Sign up", new_registration_path(resource_name) %><br />
<% end -%>
```

from that file

PR #41, opened 2017-06-07
Use a real org in the task page

rossjones:
Changes the seeds to use a real organisation (which may be updated after
organisation import) and then uses those details rather than all datasets.

### DIFF
--- a/spec/features/tasks_spec.rb
+++ b/spec/features/tasks_spec.rb
@@ -1,0 +1,47 @@
+require 'rails_helper'
+
+describe "viewing tasks" do
+  before(:each) do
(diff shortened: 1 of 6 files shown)


spec/features/tasks_spec.rb
```diff
@@ -0,0 +1,47 @@
+require 'rails_helper'
+
+describe "viewing tasks" do
+  before(:each) do
+    o = Organisation.new
```
govuklaurence: Nice - we'll probably fork out to using FactoryGirl to avoid repeating these kind of declarations, because we'll use them in many tests.

PR #57, opened 2017-06-12
Adds a utility module for link-checking, and associated rake task

rossjones:
The LinkChecker module allows for the validating of URLs to ensure
they are active and then extract as much metadata as possible from
them.  Datafiles (the thing that is actually checked) will be saved
back to the database with the new metadata, along with the datetime
that the check was carried out.

Checks can be run from the CLI using rake either at an organisational
level

```
rake linkcheck:organisation[land-registry]
```

or a dataset level

```
rake linkcheck:dataset[1862-act-registers]
```

Contains a migration to ```rake db:migrate``` AND new gems to ```bundle install```


rossjones:
Codeclimate is a little overzealous about claiming code is 'similar' and I don't have perms to tell it to shut up.

govuklaurence:
Oh, it still works, just needed approving

PR #59, opened 2017-06-13
Adds sort on tasks page

### LINKED ISSUE #49: Re-enable sorting on tasks page

hannako:
Closes #49


app/helpers/tasks_helper.rb
```diff
@@ -1,2 +1,63 @@
 module TasksHelper
+
```
govuklaurence: This could be refactored, but for now it's only complaining about duplication... we'll come back to it later

hannako: am mid refactor, one sec

hannako: This is still pretty terrible, sorry.

govuklaurence:
Will pair with you tomorrow, we'll see what we can do

PR #62, opened 2017-06-14
New Datasets Wizard

### LINKED ISSUE #54: Dataset form flow needs implementing
This includes adding datafiles.

govuklaurence:
Closes #54


app/controllers/datasets_controller.rb
```diff
@@ -1,2 +1,107 @@
...
+
+  def new
+    @dataset = Dataset.new
+  end
+
+  def create
+    @dataset = Dataset.new(params.require(:dataset).permit(:id, :title, :summary, :description))
+    @dataset.organisation = current_user.primary_organisation
+
+    if @dataset.save
```
rossjones: We should probably slugify the title into the dataset.name field here as the end url will be /dataset/NAME

app/controllers/datasets_controller.rb
```diff
@@ -1,2 +1,107 @@
...
+  def addfile
+    @dataset = current_dataset
+    @datafile = Datafile.new
+
+    if request.post?
+      file_params = params.require(:datafile).permit(:url, :name)
+      @datafile = Datafile.new(file_params)
+      @datafile.dataset = @dataset
+
+      if @datafile.save
```
rossjones: When we save a datafile, I guess this is for the model, we need to set the start_date,end_date fields based on any date data we got from the form (like month/year, or q/year).

app/models/dataset.rb
```diff
@@ -1,3 +1,19 @@
 class Dataset < ApplicationRecord
   belongs_to :organisation
+  has_many :datafiles
+
+  validates :title, presence: true, uniqueness: true
+  validates :summary, presence: true
```
rossjones: Before we publish, we also need to validate the presence of:
1. A licence  (licence or licence_other).
2. A frequency (from a fixed list).
3. At least one datafile

app/controllers/datasets_controller.rb
govuklaurence: agreed. this was just getting it working ASAP :) let's spin it out into an issue

app/controllers/datasets_controller.rb
govuklaurence: Understood -- I'll include this in a seperate issue with the changes to frequency

app/models/dataset.rb
govuklaurence: Spinning out

govuklaurence: #82

app/controllers/datasets_controller.rb
govuklaurence: #81

app/controllers/datasets_controller.rb
govuklaurence: #80

Issue #50, opened 2017-06-09
Tasks and Broken Datasets logic needs implementing


hannako:
Currently the Tasks page is displaying all tasks and all datasets.

rossjones:
I've tweaked it a bit so that it only shows ones for the org/user because when I ran an import of the legacy data, I ended up with hundreds.

We definitely need to be able to generate actual tasks, was mentioned in planning yesterday, we can start by copying the logic from alpha, but would need to work out the best way of running regular background tasks (or running a separate service for task generation, broken link checking) /cc @govuklaurence

rossjones:
The steps taken by HDX at https://github.com/OCHA-DAP/hdx-data-freshness seem like a good starting point for dealing with overdue datasets.  Let's use those steps for file-checking as a starting point?

rossjones:
PR at https://github.com/datagovuk/publish_data_beta/pull/64

PR #64, opened 2017-06-14
Overdue tasks

rossjones:
Provides a rask task which can check (at an organisation or at a dataset level
) for datasets with a frequency that are overdue.

If the frequency is not set, or a task for the dataset already exists, then
it will ignore that dataset.  Otherwise it will find the end-date for the latest
datafile and check if the dataset is missing a more recent entry.

Will require changes/refactor once model associations are done, and we still need
to find a way to do this for the entire catalogue on a schedule (maybe only weekly
for overdue datasets).

part of #50

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,22 @@ $ rails s
 $ rake db:seed
 ```
 
+## Generating tasks
+
+### Checking for broken links
+
+```
+rake linkcheck:organisation[org-short-name]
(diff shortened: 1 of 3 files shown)


rossjones:
Please ignore codeclimate, it's over-reacting.  Plus next PR will be a refactor once DB associations done.

PR #76, opened 2017-06-16
Refactor the overdue and link-check tasks into one command

### LINKED ISSUE #67: Refactor task generation code
The code for task generation repeats itself quite a lot, and could be cleaned up in such a way as to be re-usable within the app at runtime (if necessary).

rossjones:
To simplify the rake tasks, merge the two current checking tasks into
a single 'check' task that can be used as:

```
rake check:links:organisation[org-short-name]
rake check:overdue:dataset[dataset-short-name]
```

This closes #67


rossjones:
Codeclimate duplication errors are ignorable.

Issue #73, opened 2017-06-15
Form styles for dataset creation/edit


govuklaurence:

rossjones:
This was merged in https://github.com/datagovuk/publish_data_beta/pull/86

PR #86, opened 2017-06-18
Improves styling on dataset forms

hannako:
connected to #73 

- This PR adds missing class on text fields
- Nests the radio input within the radio label to fix styling.

Remaining issue: radio button styles only picked up if you land on the page after pressing skip on the previous page. If you've landed here from clicking submit (ie via Rails redirect_to) the styles aren't picked up. I think there is an issue with the js running on the page, maybe it's the order in which the assets are compiled? I don't know - I've not been able to fix it.


govuklaurence:
Spun issue out to #87

PR #90, opened 2017-06-19
Generates and uses slugs for datasets

### LINKED ISSUE #80: Use slugs instead of IDs for datasets

rossjones:
Datasets need to be available via a friendly url so we are using the
friendly_id gem to generate the unique slugs for each dataset into the
name field.  This included removing the rails unique constraint on the
title.

Requires a ```bundle install```, no migration.

Closes #80


govuklaurence:
CC issue is a critical one

rossjones:
@govuklaurence This is ready to go again now.

PR #94, opened 2017-06-19
Location javascript

maxf:
Fixes the template of the location screen, and add the javascript properly.
There are still some layout issues but that'll be dealt with as part of a general style fixing task.


.codeclimate.yml
```diff
@@ -0,0 +1,21 @@
+engines:
+  rubocop:
+    enabled: true
+  eslint:
+    enabled: true
+  csslint:
+    enabled: true
+  duplication:
```
govuklaurence: I want to re-enable this for ruby later

Gemfile
```diff
@@ -19,6 +19,7 @@ gem 'kaminari', '~> 1.0.1'
 gem 'rest-client', '~> 2.0.2'
 gem 'mime-types', '~> 3.1'
 gem 'friendly_id', '~> 5.2.1'
+gem 'jquery'
```
govuklaurence: I don't think this is needed - jquery-rails should cover it

app/assets/javascripts/main.js
```diff
@@ -0,0 +1,191 @@
+/* eslint-disable no-unused-expressions */
+/* global $, Locations */
+
+;(function () {
```
govuklaurence: Semicolon is not required

rossjones: It's a safety fallback for when we concatenate multiple JS files. It's fairly common practice I think.

Gemfile
rossjones: Seems to work without gem 'jquery' but I've no locations in my DB (how do I do that?).

I do get this from my console
```
Started GET "/api/locations?q=liverpoo" for 127.0.0.1 at 2017-06-20 09:29:45 +0100
Processing by LocationsController#lookup as JSON
  Parameters: {"q"=>"liverpoo"}
  User Load (0.6ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 ORDER BY "users"."id" ASC LIMIT $2  [["id", 1], ["LIMIT", 1]]
  Location Load (0.3ms)  SELECT "locations".* FROM "locations" WHERE (name ILIKE CONCAT('%','liverpoo','%'))
Completed 200 OK in 4ms (Views: 0.1ms | ActiveRecord: 0.9ms)
```

PR #119, opened 2017-06-27
Editing Datasets and a full suite of tests

govuklaurence:
This is huge -- please take your time over this review


app/controllers/datasets_controller.rb
```diff
@@ -13,54 +23,75 @@ def create
...
     @dataset = current_dataset
 
-    if request.post?
+    unless request.get?
       licence = get_licence(params.require(:dataset).permit(:licence, :licence_other))
       @dataset.licence = licence
 
-      redirect_to new_location_dataset_path(@dataset) if @dataset.save
+      if @dataset.save
+        redirect_to new_location_dataset_path(@dataset) if request.post?
```
rossjones: I'm guessing both redirects won't work

app/controllers/datasets_controller.rb
```diff
@@ -13,54 +23,75 @@ def create
...
     @dataset = current_dataset
 
-    if request.post?
+    unless request.get?
       location_params = params.require(:dataset).permit(:location1, :location2, :location3)
       @dataset.update_attributes(location_params)
 
-      redirect_to new_frequency_dataset_path(@dataset) if @dataset.save
+      if @dataset.save
+        redirect_to new_frequency_dataset_path(@dataset) if request.post?
```
rossjones: Two redirects?

config/routes.rb
```diff
@@ -13,23 +13,23 @@
...
 
+    member do
+      match 'licence',       to: 'datasets#licence',     via: [:post, :put]
+      match 'location',      to: 'datasets#location',    via: [:post, :put]
+      match 'frequency',     to: 'datasets#frequency',   via: [:post, :put]
+
+      resources :files,     controller: 'datafiles', param: :file_id
+      resources :documents, controller: 'datafiles', param: :file_id
+
+      get 'new/licence',    to: 'datasets#licence'
```
rossjones: Not a big fan of new/ and edit/ although I know it makes it easier to determine whether you're editing or not, but I guess it doesn't matter too much.  Maybe we can discuss it next time we're together to see if it introduces more complexity or less.  Just interested in what the idiomatic ruby way is.

app/helpers/datasets_helper.rb
```diff
@@ -8,6 +8,22 @@ def dataset_field(f, dataset, options)
   }
 
   def friendly_frequency(frequency)
-    FRIENDLY_FREQUENCIES.fetch(frequency, 'One-off')
+    FRIENDLY_FREQUENCIES.fetch(frequency, frequency.humanize)
+  end
```
maxf: Probably ok for now, but humanize won't do it as soon as we have content changes. It also doesn't work with internationalisation/localisation.

app/views/datafiles/_dataset_field.html.erb
```diff
@@ -0,0 +1,39 @@
+<div class="form-group<%= " error" if dataset.errors.include?(field[:name])%>" id="<%= field[:name] %>-form-group">
```
maxf: this fill will probably conflict with #116 .
Also, I'm a bit confused by this file as it seems to be a generic partial for forms controls, which is great but other templates aren't using it systematically.

(which reminds me I'm considering using https://github.com/ministryofjustice/govuk_elements_form_builder as a possible improvement to govuk forms)

app/controllers/datasets_controller.rb
```diff
@@ -13,54 +23,75 @@ def create
...
+      end
     end
   end
 
   def frequency
     @dataset = current_dataset
 
-    if request.post?
+    unless request.get?
       @dataset.frequency = params.require(:dataset).permit(:frequency)[:frequency]
```
hannako: @govuklaurence if you skip selecting the frequency params[:dataset] is empty

screen shot 2017-06-28 at 10 34 37

rossjones: Shouldn't be able to skip selecting a frequency, so might be easier to fix that way for now.

app/assets/javascripts/datafiles.coffee
```diff
@@ -0,0 +1,3 @@
+# Place all the behaviors and hooks related to the matching controller here.
+# All this logic will automatically be available in application.js.
+# You can use CoffeeScript in this file: http://coffeescript.org/
```
govuklaurence: it isn't, will kill it

app/assets/stylesheets/datafiles.scss
```diff
@@ -0,0 +1,3 @@
+// Place all the styles related to the datafiles controller here.
+// They will automatically be included in application.css.
+// You can use Sass (SCSS) here: http://sass-lang.com/
```
govuklaurence: will kill

app/controllers/datasets_controller.rb
govuklaurence: multiple redirects throws an exception

govuklaurence: really what i should do is seperate these into four methods (new, edit, create, update)

app/controllers/datasets_controller.rb
govuklaurence: same as prior

app/controllers/datasets_controller.rb
govuklaurence: will add a validation here

app/helpers/datasets_helper.rb
govuklaurence: going to leave this as it is for now, it's an easy fix to change it

app/views/datafiles/_dataset_field.html.erb
govuklaurence: Aye, this will quite possibly have to change

config/routes.rb
govuklaurence: We should probably talk this over on thursday. I'm quite sure there's a far better way to do this. This is the result of a pile of hacks.

app/controllers/datafiles_controller.rb
```diff
@@ -0,0 +1,172 @@
...
+    :year,
+    :start_day,
+    :start_month,
+    :start_year,
+    :end_day,
+    :end_month,
+    :end_year
+  ]
+
+  def set_dates(date_params)
```
hannako: @govuklaurence If no date is entered when you are adding a link to a datafile there's an error...

screen shot 2017-06-28 at 10 55 38

app/views/datafiles/files.html.erb
```diff
@@ -0,0 +1,39 @@
...
+      </tr>
+    </thead>
+  <tbody>
+    <% @datafiles.each do |link| %>
+      <tr id="file_1">
+        <th><a href="<%= link.url %>"><%= link.name %></a></th>
+        <td>✔</td>
+        <td class="actions">
+          <%= link_to 'Edit', edit_file_path(@dataset.id, link.id) %>
+          <%= link_to 'Delete', file_path(@dataset.id, link.id), method: :delete %>
```
hannako: I think you can just pass in `@dataset`, I thought it was bad juju to do `@dataset.id`

govuklaurence: you're correct

app/controllers/datafiles_controller.rb
govuklaurence: understood, will add a test case

app/views/datafiles/_date_field.html.erb
```diff
@@ -0,0 +1,27 @@
...
+      <span class="form-label-bold">
+          <%= period.humanize %> Date
+          <span class="form-hint">For example, 31 3 2016</span>
+      </span>
+    </legend>
+
+    <div class="form-date">
+      <div class="form-group form-group-day">
+        <label for="#<%= period %>_day">Day</label>
+        <%= text_field_tag "datafile[#{period}_day]", nil, id: "#{period}_day", class: "form-control" %>
```
hannako: Just asking the question if it's ok that this is `nil`, i.e. that when you're on the edit page everything is auto-populated other than the date...

govuklaurence: that's correct, thanks for pointing that out... i should replace that.

app/views/datasets/dataset.html.erb
```diff
@@ -22,7 +22,8 @@
 		</div>
   <% end %>
 
-  <%= form_for @dataset, action: 'create', method: 'post' do |f| %>
+  <%= form_for @dataset, url: editing? ? dataset_path(@dataset) : datasets_path,
+                         method: update_method do |f| %>
```
hannako: Something funny around here. If you try to submit the form without a required field the url changes to `/datasets` which means you can't then submit a post request.

app/controllers/datafiles_controller.rb
hannako: Should this be a new issue as it's not fixed

app/views/datafiles/_date_field.html.erb
hannako: should this be a new issue as it's still bust

app/controllers/datafiles_controller.rb
govuklaurence: #128

app/views/datafiles/_dataset_field.html.erb
govuklaurence: #129

app/views/datafiles/_date_field.html.erb
govuklaurence: #130

govuklaurence:
Populating date fields - #133

PR #127, opened 2017-06-30
Frontend refactor

maxf:


maxf:
@hannako : no technical reason, just not got round to doing it, and this PR is already quite big. More fixes to be made later though, as discussed IRL.

PR #137, opened 2017-07-03
Fix Datafile form prefill

govuklaurence:
When a datafile is being edited, it should show any existing dates that have been entered. This change adds that, as well as fixing a display bug.


govuklaurence:
Fixes #138

app/views/datafiles/_quarter_inputs.html.erb
```diff
@@ -9,26 +9,26 @@
 		
 		<div class="form-group " id="quarter_form_group">
 			<label for="q1" class="block-label selection-button-radio">
```
maxf: tabs

govuklaurence: your wish is vim's command...

PR #141, opened 2017-07-03
Use env var to specify location of FIND_URL

rossjones:


govuklaurence:
We should add a helper method for generating find urls

PR #153, opened 2017-07-04
Improve coverage of organisation model

rossjones:


govuklaurence:
CC should be fairly easy to dedupe for this one

rossjones:
We probably shouldn't use a sequence for organisation, if the name is taken then we're obv trying to add a duplicate org and we don't want that.

PR #169, opened 2017-07-07
Elastic client setup at rails start

rossjones:


govuklaurence:
should move the elastic config to a yml file that is pulled by the initializer

rossjones:
Will add a ticket.

PR #176, opened 2017-07-10
Add instructions

maxf:


spec/end-to-end-tests/run_fe_tests.sh
```diff
@@ -46,7 +46,8 @@ rake db:seed > /dev/null 2>&1
 rake import:locations\['lib/seeds/locations.csv'\]
 
 echo Starting test server
-rails s -p ${TEST_APP_PORT} >/dev/null 2>&1 &
+rm -f tmp/pids/e2e-server.pid
```
govuklaurence: base off this file location -- if this is run from another dir it may do something unintended

PR #178, opened 2017-07-10
Only allow publishing if the dataset has >0 links

rossjones:


govuklaurence:
Already doing a similar change

PR #187, opened 2017-07-10
Fix a couple of crashes on creation

govuklaurence:
* No frequency entered
* Documentation with month selected


app/controllers/datasets/frequencies_controller.rb
```diff
@@ -11,7 +11,14 @@ def edit
...
     @dataset = current_dataset
-    @dataset.frequency = params.require(:dataset).permit(:frequency)[:frequency]
+
+    begin
+      @dataset.frequency = params.require(:dataset).permit(:frequency)[:frequency]
+    rescue ActionController::ParameterMissing
+      @dataset.errors.add(:frequency, 'No frequency selected')
+      render 'new'
+      return
+    end
```
maxf: Is it ok to use exceptions in controllers? Seems odd to me, but maybe it's standard practice in Rails

hannako: I think maybe you can use `rescue_from with :handler` http://api.rubyonrails.org/classes/ActiveSupport/Rescuable/ClassMethods.html

govuklaurence: It's a fairly normal and benign usage on a very specific error. Don't want to use a handler because it would split up the flow.

spec/end-to-end-tests/tests/create.js
```diff
@@ -82,18 +82,12 @@ var test_create_happy_path = function (browser) {
...
     .clearSetValue('input[id=id_url]', common.validDataUrl)
     .clearSetValue('input[id=id_name]', common.docTitle)
-    .saveScreenshot('wtfbbq4.jpg')
     .submitFormAndCheckNextTitle('Links to supporting documents')
-    .saveScreenshot('wtfbbq5.jpg')
     .clickAndCheckNextTitle('Save and continue', 'Publish ‘' + common.datasetTitle  + '’')
-    .saveScreenshot('wtfbbq6.jpg')
     .assert.containsText('table', 'Open Government Licence')
     .submitFormAndCheckNextTitle('Your dataset has been published')
     .end();
```
maxf: Next time can you rebase so we don't have those extra commits?

app/models/datafile.rb
```diff
@@ -71,10 +71,17 @@ def dates
...
   end
 
+  def document?
+    documentation == true
+  end
+
+  def file?
+    !document?
+  end
+
```
maxf: Are those functions used anywhere?

govuklaurence: They are left over from another fi that ross already did, will remove.

Issue #180, opened 2017-07-10
Confirm delete link process


govuklaurence:
This is present in alpha but not in beta.

Repro:
1. Navigate to alpha
2. Log in
3. Manage Datasets
4. New Dataset
5. Add an example link
6. On publish page, edit files
7. Delete
8. Confirm
9. File is deleted

This process needs to be copied over

maxf:
Part of #179, which also covers delete datasets

Issue #179, opened 2017-07-10
Add confirmation for delete dataset or file


maxf:
When deleting a file or a dataset there should be a confirmaiton dialog

PR #204, opened 2017-07-11
Delete datasets

hannako:
Datasets can be deleted

- let me know if this looks ok and I will apply same to deletion of datafiles. 

relates to #180 and #179

### DIFF
--- a/app/controllers/datasets_controller.rb
+++ b/app/controllers/datasets_controller.rb
@@ -105,6 +105,18 @@ class DatasetsController < ApplicationController
     end
   end
 
+  def confirm_delete
+    @dataset = current_dataset
+    flash[:confirm_delete] = 'Are you sure you want to delete this dataset?'
+    render 'show'
+  end
+
+  def destroy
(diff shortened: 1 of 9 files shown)


app/controllers/datasets_controller.rb
```diff
@@ -105,6 +105,29 @@ def publish
     end
   end
 
+  def confirm_delete
+    @dataset = current_dataset
+    if @dataset.published?
+      @dataset.errors.add(:delete_prevent, 'Published datasets cannot be deleted')
+    else
+      flash[:confirm_delete] = 'Are you sure you want to delete this dataset?'
```
govuklaurence: You could use `flash[:alert]` or something of that ilk, i believe rails has a default for it.

app/controllers/datasets_controller.rb
```diff
@@ -105,6 +105,29 @@ def publish
...
+    end
+    render 'show'
+  end
+
+  def destroy
+    @dataset = current_dataset
+    if @dataset.published?
```
govuklaurence: might be worth preventing in the model as well

PR #248, opened 2017-07-17
Use correct value for disabling autocomplete

maxf:


spec/end-to-end-tests/tests/create.js
```diff
@@ -182,19 +182,6 @@ var test_create_add_file_twice = function (browser) {
 module.exports = {
   'Create a dataset, happy path': test_create_happy_path,
 //  'Create a dataset, invalid title': test_create_invalid_title,
```
hannako: should this be deleted instead of commented out?

spec/end-to-end-tests/tests/create.js
```diff
@@ -182,19 +182,6 @@ var test_create_add_file_twice = function (browser) {
...
-  'Create a dataset, region autocomplete': test_create_region_autocomplete,
```
hannako: this test and some of the next deleted ones are still defined above. Are you keeping them there in case you want to run them again in the future?

spec/end-to-end-tests/tests/create.js
maxf: Yes, I'll remove it. I'd left it there to remind me about discussing what's a valid dataset name, but I'll record that as an issue instead.

spec/end-to-end-tests/tests/create.js
maxf: no, it's just me being overzealous

Issue #183, opened 2017-07-10
Refactor datafiles controller and views


hannako:
split out files and documents

PR #250, opened 2017-07-17
Simplify datafile models to link and doc

### LINKED ISSUE #184: Simplify Datafile model
- make datafile object polymorphic, so it can be either a document or a file

govuklaurence:
Fixes #184

THIS IS NOT REPRESENTATIVE OF THE FINISHED STATE OF #183! IT IS AN INTERMEDIATE PULL THAT REFACTORS THE MODELS ONLY.

Migrations:
* One column added to datafiles table

Model Changes:
* Docs and Links added, children of Datafile

Other Changes:
* Date Validator renamed and moved to `lib`


govuklaurence:
CC Issues are WONTFIX for the time being since they are old issues

PR #254, opened 2017-07-18
Imports changed datasets from live where change <24 hours ago.

### LINKED ISSUE #236: Daily import
We should build a task that can be run every X hours that will import any legacy datasets that have changed in the last X hours.  We should refuse to import if it has been changed on publish as well.

rossjones:
Adds a rake task to import datasets which have been modified in the
last 24 hours (by asking live DGU for the metadata).

Fixes #236

### DIFF
--- a/lib/util/metadata_tools.rb
+++ b/lib/util/metadata_tools.rb
@@ -1,0 +1,184 @@
+module MetadataTools
+
+  def add_dataset_metadata(obj, orgs_cache)
+    d = Dataset.find_or_create_by(uuid: obj["id"])
+    d.name = obj["name"]
+    d.title = obj["title"]
+    d.summary = generate_summary(obj["notes"])
+    d.description = obj["notes"]
+    d.organisation_id = orgs_cache[obj["owner_org"]]
+    d.licence = obj["license_id"]
+    d.licence = 'no-licence' if d.licence == ""
+    d.licence_other = ""
+    #d.frequency = convert_frequency(obj["update_frequency"])
+    d.frequency = "never"
+    d.published = false
+    d.published_date = obj["metadata_created"]
+    d.created_at = obj["metadata_created"]
+    d.updated_at = obj["metadata_modified"]
+    d.dataset_type = dataset_type(obj)
+    d.harvested = harvested?(obj)
+    d.location1 = ""
+    d.location2 = ""
+    d.location3 = ""
+    d.legacy_metadata = ""
+    d.save!()
+
+    # Add the inspire metadata if we have determined this is a ULKP
+    # dataset.
+    if d.dataset_type == 'inspire'
+      inspire = add_inspire_metadata(d.id, obj)
+      inspire.save!()
+    end
+
+    # Iterate over the resources list and add a new datafile for each
+    # item.
+    obj["resources"].each do |resource|
+      if get_doc_type(resource['format'])
+        datafile = Doc.find_or_create_by(url: resource["url"], dataset_id: d.id)
+      else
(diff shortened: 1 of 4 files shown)


lib/util/metadata_tools.rb
```diff
@@ -0,0 +1,187 @@
...
+    d = Dataset.find_or_create_by(uuid: obj["id"])
+    d.name = obj["name"]
+    d.title = obj["title"]
+    d.summary = generate_summary(obj["notes"])
+    d.description = obj["notes"]
+    d.organisation_id = orgs_cache[obj["owner_org"]]
+    d.licence = obj["license_id"]
+    d.licence = 'no-licence' if d.licence == ""
+    d.licence_other = ""
+    #d.frequency = convert_frequency(obj["update_frequency"])
```
hannako: is this mean to be commented out, and is setting them all to `never` right?

lib/util/metadata_tools.rb
```diff
@@ -0,0 +1,187 @@
...
+  def add_dataset_metadata(obj, orgs_cache)
+    d = Dataset.find_or_create_by(uuid: obj["id"])
+    d.name = obj["name"]
+    d.title = obj["title"]
+    d.summary = generate_summary(obj["notes"])
+    d.description = obj["notes"]
+    d.organisation_id = orgs_cache[obj["owner_org"]]
+    d.licence = obj["license_id"]
+    d.licence = 'no-licence' if d.licence == ""
+    d.licence_other = ""
```
govuklaurence: ^ agreed

rossjones: There's a diff between no-licence and other :) We don't know what it is if it isn't specified so can't set to other.  This does need further work though for other licences we do know about.

lib/util/metadata_tools.rb
```diff
@@ -0,0 +1,187 @@
...
+      inspire.save!()
+    end
+
+    # Iterate over the resources list and add a new datafile for each
+    # item.
+    obj["resources"].each do |resource|
+      add_resource(resource, d)
+    end
+
+    d.published = true
```
govuklaurence: i think it's assuming that everything in the import IS published, which may be true - @rossjones ?

lib/util/metadata_tools.rb
rossjones: Yes, it's not right, we need to do more in another PR.

lib/util/metadata_tools.rb
```diff
@@ -0,0 +1,187 @@
...
+    inspire.spatial_data_service_type = get_extra(extras, 'spatial-data-service-type')
+    inspire.spatial_reference_system = get_extra(extras, 'spatial-reference-system')
+    inspire
+  end
+
+  # Generates a summary from the provided notes as in
+  # the legacy metadata we don't have a summary field.
+  def generate_summary(notes)
+    return notes if notes && notes != ""
+    "No description provided"
```
rossjones: It's handled by the view, but we could also do it here.

lib/util/metadata_tools.rb
```diff
@@ -0,0 +1,187 @@
...
+  end
+
+  # Returns a value if the dataset is harvested, i.e. it has
+  # a harvest object id.
+  def harvested?(obj)
+    get_extra(obj["extras"], "harvest_object_id") != ""
+  end
+
+  # Determine whether datafile is documentation or not
+  def get_doc_type(fmt)
```
rossjones: Yeah, will fix in the next cleanup PR.

PR #257, opened 2017-07-18
Maxf/error styling

maxf:


spec/features/create_dataset_spec.rb
```diff
@@ -178,7 +178,6 @@
 
         expect(page).to     have_content('Start Date')
         expect(page).to     have_content('End Date')
-        expect(page).to_not have_content('Year')
```
hannako: why did this get deleted?

app/views/datasets/show.html.erb
```diff
@@ -17,8 +17,8 @@
     <% else %>
     <div id="confirm-delete-box">
       <h1 class="heading-medium error-summary-heading"><%= flash[:alert] %></h1>
-        <%= link_to "Yes, delete this dataset", @dataset, method: :delete, class: 'button confirm-delete-box__button' %>
```
hannako: thanks for tidying up my indentation!

spec/features/create_dataset_spec.rb
maxf: Because they shouldn't have been there in the first place (Year should be present on all datafile forms (except daily and one-off). However the test still passed because the template had 'year' instead of 'Year'. When I corrected the capitalisation, that test started to fail, because of the test itself.

PR #262, opened 2017-07-19
Assign organisations to hierarchy when loading

### LINKED ISSUE #252: Fix organisation import
Fix the organisation import so that it creates the organisations as a hierarchy.

rossjones:
When loading organisations, will perform a second stage where the
organisation is assigned to the correct parent.

Fixes #252

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -16,6 +16,8 @@ namespace :import do
   task :organisations, [:filename] => :environment do |_, args|
     count = 1
 
+    relationships = {}
+
     json_from_lines(args.filename) do |obj|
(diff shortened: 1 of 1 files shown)


lib/tasks/import.rake
```diff
@@ -34,11 +36,28 @@ namespace :import do
...
     end
+
+    puts "Processing #{relationships.size} child organisations"
+
+    count = 0
+    relationships.each do |child, parent|
+      o = Organisation.find_by(name: child)
+      o.parent = Organisation.find_by(name: parent)
+      o.save!()
+      print "Assigned #{count+=1} organisations...\r"
```
govuklaurence: thank god i think you're the only other person i've seen use carriage returns. thank you.

PR #269, opened 2017-07-19
Creates a data.json file for all published datasets

### LINKED ISSUE #124: Generate a data.json file
Publish Data (as it has all of the metadata) should generate a data.json file (on a schedule) that can be hosted somewhere but made available for Find-Data at https://..../data.json 

This will allow other catalogues to harvest ours without hammering an API as it does currently.

rossjones:
Once the file is created, it is moved to public/ so that it can be served
by publish data. This would require find-data to route to the right place
but in future we might store this file on S3.

Fixes #124


lib/tasks/generate.rake
```diff
@@ -0,0 +1,85 @@
...
+      print "Encoded #{count+=1} datasets...\r"
+    end
+
+    enum.yield "\n]}"
+  end
+end
+
+def dataset_record(dataset)
+  record = {
+    "@type"      => "dcat:Dataset",
```
govuklaurence: https://github.com/bbatsov/ruby-style-guide#symbols-as-keys

lib/tasks/generate.rake
```diff
@@ -0,0 +1,85 @@
...
+    resources << resource
+  end
+
+  record["distribution"] = resources
+
+  "\n#{record.to_json}"
+end
+
+def licence_for_id(id)
+  return "http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" if id == "uk-ogl"
```
govuklaurence: i would nix the return and make this a ternary, but it's equivalent so shrug

lib/tasks/generate.rake
```diff
@@ -0,0 +1,85 @@
...
+
+    file.close()
+    File.rename "data.json", "public/data.json"
+
+  end
+
+end
+
+# Generates a string to write to the output file
+def generate_string()
```
govuklaurence: what is this method doing? quite confused by the use of enumerator here

lib/tasks/generate.rake
```diff
@@ -0,0 +1,85 @@
...
+
+    generate_string().each do | blob |
+      file.puts blob
+    end
+
+    file.close()
+    File.rename "data.json", "public/data.json"
+
+  end
+
```
govuklaurence: you newline maverick

rossjones: :( Will fix.

lib/tasks/generate.rake
rossjones: Rather than generate a 90Mb JSON file in memory (along with the associated models and extra data) this method yields data as it is called so that we've only got one map and one json blob in ram at any one time.  May be slightly slower, but it's the closest to a stream of json objects to be written that we can get.

lib/tasks/generate.rake
rossjones: Yeah, can't shake the pythonisms

lib/tasks/generate.rake
govuklaurence: Alright, sounds good. Just wasn't immediately clear from reading.

(That said, 90Mb really isn't that much)

lib/tasks/generate.rake
govuklaurence: It's no worries, still works

rossjones:
This is ready to go again now.

PR #270, opened 2017-07-19
[261] - Move model validation errors to locales

TobyRet:
Link to ticket:
https://waffle.io/datagovuk/publish_data_beta/cards/596f445968ba73003ebd576b

(Will do application text in another PR)

### DIFF
--- a/app/models/datafile.rb
+++ b/app/models/datafile.rb
@@ -1,6 +1,6 @@
 class Datafile < ApplicationRecord
   belongs_to :dataset
 
-  validates :url, presence: { message: 'Please enter a valid URL' }
-  validates :name, presence: { message: 'Please enter a valid name' }
+  validates_presence_of :url
+  validates_presence_of :name
 end
--- a/app/models/datafile.rb
+++ b/app/models/datafile.rb
@@ -1,6 +1,6 @@
 class Datafile < ApplicationRecord
   belongs_to :dataset
 
-  validates_presence_of :url
-  validates_presence_of :name
+  validates :url, presence: true
+  validates :name, presence: true
 end
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -1,6 +1,7 @@
 require 'elasticsearch/model'
 
 class Dataset < ApplicationRecord
+  TITLE_FORMAT = /([a-z]){3}.*/i
   include Elasticsearch::Model
   extend FriendlyId
   before_destroy :prevent_if_published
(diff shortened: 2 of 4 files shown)


config/locales/en.yml
```diff
@@ -16,3 +16,30 @@
 en:
   dataset_published: "Your dataset has been published"
   dataset_updated: "Your dataset has been edited"
+  activerecord:
```
govuklaurence: this file looks good.

app/models/datafile.rb
```diff
@@ -1,6 +1,6 @@
 class Datafile < ApplicationRecord
   belongs_to :dataset
 
-  validates :url, presence: { message: 'Please enter a valid URL' }
-  validates :name, presence: { message: 'Please enter a valid name' }
+  validates_presence_of :url
```
govuklaurence: This syntax is deprecated in rails 5 - `validates` should be used everywhere.

TobyRet:
Sorry, methinks I don't understand Waffle.
[261] is supposed to be a reference to the ticket number

govuklaurence:
You can use regular old github issue links, e.g. #261 :)

app/models/dataset.rb
```diff
@@ -17,36 +18,16 @@ class Dataset < ApplicationRecord
...
-    presence: { message: "Please select a licence for your dataset" },
-    if: lambda{ published }
-
-  validates :licence_other,
-    presence: { message: "Please type the name of your licence" },
-    allow_blank: false,
-    if: lambda { licence == 'other' }
-
+  validates_presence_of :title
+  validates_format_of :title, :with => TITLE_FORMAT
```
govuklaurence: `:symbol => value` should always be written as `symbol: value` 

https://github.com/bbatsov/ruby-style-guide#symbols-as-keys

config/locales/en.yml
```diff
@@ -16,3 +16,30 @@
...
+        dataset:
+          attributes:
+            title:
+              invalid: "Please enter a valid title"
+              blank: "Please enter a valid title"
+            summary:
+              blank: "Please provide a summary"
+            frequency:
+              blank: "Please indicate how often this dataset is updated"
+              inclusion: ['daily', 'weekly', 'monthly', 'quarterly', 'annually', 'financial-year', 'never']
```
govuklaurence: This will set an incorrect message - it's a property of the inclusion validator and needs to stay on the model

app/models/dataset.rb
```diff
@@ -17,36 +18,15 @@ class Dataset < ApplicationRecord
 
   friendly_id :slug_candidates, :use => :slugged, :slug_column => :name
 
-  validates :frequency, inclusion: %w(daily weekly monthly quarterly annually financial-year never),
+  validates :frequency, inclusion: {in: %w(daily weekly monthly quarterly annually financial-year never)},
```
govuklaurence: linter may complain about spaces near braces

PR #274, opened 2017-07-19
[DO NOT MERGE] Split out docs' and links' controllers and views

govuklaurence:
* Split views into two folders
* Split datafiles controller in two
* Fix tests
* Break creation flow (does not go to docs page)

TODO:

- [x] Refactor views
- [x] Delete datafile controller, views, etc
- [x] Un-break creation flow


app/views/docs/_date_field.html.erb
```diff
@@ -0,0 +1,41 @@
+<div class="form-group" id="<%= period %>_date_form_group">
```
rossjones: Does this make sense at all for docs? They don't have associated dates

app/views/docs/_month_input.html.erb
```diff
@@ -8,20 +8,20 @@
     </legend>
 
     <div class="form-date">
-      <div class="form-group form-group-month<% if @datafile.errors[:start_month].present? %> form-group-error<% end %>">
+      <div class="form-group form-group-month<% if @doc.errors[:start_month].present? %> form-group-error<% end %>">
```
rossjones: No dates for docs

app/views/docs/_date_field.html.erb
govuklaurence: Ah, that's just leftovers. will remove.

PR #281, opened 2017-07-21
Validate data link urls

TobyRet:
Created a new ActiveModel Custom validator to check data link urls.
See issue @ https://waffle.io/datagovuk/publish_data_beta/cards/59538ecba67fad0354d4a004


lib/validators/url_validator.rb
```diff
@@ -0,0 +1,56 @@
...
+      RestClient.head record.url
+      true
+    rescue RestClient::ExceptionWithResponse
+      error = 'Url path is not valid'
+      createValidationError(record, error)
+    end
+  end
+
+  def createValidationError(record, error)
+    Rails.logger.info('Validation error: ' + error)
```
TobyRet: Currently we are flashing the same error message regardless of why the url validation failed. I think we need to wait confirmation of content. In the meantime thought it might be useful to log the exact error.

govuklaurence: `logger.debug` is probably the way to go

PR #285, opened 2017-07-24
Adds a temporary page to view metadata quality.

rossjones:
Temporarily adds a new page to allow us to see why we think a dataset
is bad quality, it is at /datasets/NAME/quality.  We may (or not) want
to integrate something similar at some point so that publishers know
why we consider a dataset bad quality.

New scores can be added to the class at /lib/quality/quality_score
by adding a new method that ends with _score, it should return an integer.

This is to close https://trello.com/c/Dbs1tDIg/226-quality-reporting-in-publish

### DIFF
--- a/lib/quality/quality_score.rb
+++ b/lib/quality/quality_score.rb
@@ -42,11 +42,11 @@ class QualityScore
 
     # Are there any links at all?
     current = if links.size() == 0
-      @reasons << "There are no data links in this dataset"
-      10
-    else
-      0
-    end
+                @reasons << "There are no data links in this dataset"
+                10
+              else
+                0
+              end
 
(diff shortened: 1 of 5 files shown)


lib/quality/quality_score.rb
```diff
@@ -0,0 +1,103 @@
...
+    if @dataset.frequency.blank?
+      @reasons << "This dataset has no update frequency set"
+      10
+    else
+      0
+    end
+  end
+
+  def documentation_score
+    if @dataset.docs.count() == 0
```
hannako: Not sure you need the brackets here or in the other methods ie...`@dataset.docs.count == 0`, `@dataset.links.all` etc

PR #286, opened 2017-07-24
Autoset UUIDs where needed rather than relying on uuid-ossp

rossjones:


govuklaurence:
Might be easier to do it on ApplicationRecord... all models use UUID, right?

rossjones:
No, only some models.

PR #293, opened 2017-07-27
Remove attempt to load UUID extension

rossjones:
Otherwise migrations break on Paas.

Also adds the paas.sh to workaround PaaS not leaving bin/rails in a runnable state when you ssh in to the container.  Should run as 

```
cd app
../tools/paas.sh bin/rails something....
```


rossjones:
If someone wanted to squash the migrations, that would also be awesome...

PR #296, opened 2017-07-28
Organisation scores

rossjones:
Adds a page at /quality that lists the scores for a particular organisation.
It shows the highest score, the lowest score, and the median.

Scores have to be generated offline with the quality:calculate_all rake task.

Once this is no longer needed, it will need to be removed.


lib/tasks/quality.rake
```diff
@@ -1,13 +1,65 @@
...
   task :score,[:dataset] => :environment do |_, args|
     dataset = Dataset.find_by(name: args.dataset)
-    q = QualityScore.new(dataset)
-    p q.score
-    p q.reasons
+    q = QualityScoreCalculator.new(dataset)
   end
 
+  desc "Calculate scores for an organisation"
+  task :calculate, [:organisation] => :environment do |_, args|
```
hannako: Just so we don't have to bug you once you're gone, might be worth making a comment that it's expecting short-name or something.

PR #299, opened 2017-07-31
Tries harder to import resources with dates attached

rossjones:
As a drive-by, also sets the INSPIRE guid which was missing in earlier import


rossjones:
Messed up my merging.

PR #302, opened 2017-08-01
Adds a model and a rake task for generating previews.

rossjones:
This is the initial backend of the preview code.

Adds a new model linked to the link table, containing a json blob for
the preview of that link, if it's a CSV.

A new rake task ```preview:generate_all``` will iterate through the database
and find all links claiming to be CSVs (hint they're often not) and trying
to create a preview entry for them.

Because we don't have the faintest idea of what the encoding of the incoming CSV
is, we use the ```file``` command to find out what it is, and convert it to
UTF-8 if possible using iconv.

Still needs a UI, changes to dataset view to link to page and hooking in
to new links being created (as a sidekiq task). As preview blob is reasonably 
small, we might want to have it indexed with the link (or alternatively in a separate
index).

Requires a bundle install

Requires a migration

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -26,6 +26,7 @@ gem "ancestry", "~> 3.0.1"
 gem "govuk-lint", "~> 2.1.0"
 gem 'whois-parser'
 gem "sidekiq", "~> 5.0.4"
+gem 'iconv', "~> 1.0.4"
 
 group :development, :test do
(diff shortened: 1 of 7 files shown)


rossjones:
At current usage levels, and just for CSVs this might get to 30Mb of data in the DB.

PR #307, opened 2017-08-02
A preview API

rossjones:
When a file id is given to /api/previews/:file_id then it will return a JSON
blob containing enough information to show a preview of the file.

e.g

```json
{
  id: 1,
  link_id: 11,
  content: {
    type: "CSV",
    body: [
	 [
	  "Name:",
	  "Age"
	],
	[
	  "Bob",
	  "23"
	],
	[
	  "Fred",
	  "24"
	],
	[
	  "Jane",
	  "22"
	]
      ]
    },
    created_at: "2017-08-02T14:23:03.710Z",
    updated_at: "2017-08-02T14:23:03.710Z"
}
```

### DIFF
--- a/app/controllers/links_controller.rb
+++ b/app/controllers/links_controller.rb
@@ -1,5 +1,6 @@
 class LinksController < ApplicationController
   before_action :set_current_dataset
+  skip_before_action :set_current_dataset, only: [:preview]
 
   def new
     @link = Link.new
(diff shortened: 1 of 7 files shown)


rossjones:
To determine the encoding, we run the ```file``` command and ask it, then do the conversion to utf8 if necessary.

The mimetype isn't correctly reported by ```file``` because it comes as text/plain which it is, but not so helpful.  The current type field if the format, which was verified during the link check.  So I think we should use this solely to determine how to view it.  For now CSV will show a table, but perhaps geojson might use a map.

The filesize is recorded in the link, so will add that to the preview, but as we're only showing 5 rows I'm not sure what use the size is - it is shown (or should be) on the dataset page.

PR #311, opened 2017-08-04
Import Themes

rossjones:
When importing legacy datasets, this PR will attach a reference to an entry in the Theme
table so that we don't lose data.  We won't be displaying this anywhere right now though.

Some datasets don't have themes, in which case this will be nil.

The default themes are created in db:seed so this will need to be run for themes to be
created before import, otherwise there will be no theme references.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -17,6 +17,7 @@ class Dataset < ApplicationRecord
   before_save :set_uuid
 
   belongs_to :organisation
+  belongs_to :theme, optional: true
   has_many :links
   has_many :docs
   has_one :inspire_dataset
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -18,6 +18,8 @@ class Dataset < ApplicationRecord
 
   belongs_to :organisation
   belongs_to :theme, optional: true
+  belongs_to :secondary_theme, optional: true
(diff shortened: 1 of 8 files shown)


app/models/dataset.rb
```diff
@@ -17,6 +17,7 @@ class Dataset < ApplicationRecord
   before_save :set_uuid
 
   belongs_to :organisation
+  belongs_to :theme, optional: true
```
govuklaurence: is there a case where something can belong in multiple themes?

db/migrate/20170804132655_add_theme_field_to_dataset.rb
```diff
@@ -0,0 +1,6 @@
+class AddThemeFieldToDataset < ActiveRecord::Migration[5.1]
+  def change
+    add_column :datasets, :theme_id, :integer
+  end
+end
+
```
govuklaurence: extra space

rossjones: Fixed.

app/models/dataset.rb
rossjones: Yes. Some, as a secondary-theme.  Will add them, not sure why I forgot.

PR #326, opened 2017-08-23
Get location and last updated from legacy datasets

hannako:
- This is the back end part of the fix for this [pr](https://github.com/datagovuk/find_data_beta/pull/180). 

- We are storing the value in `metadata_modified` on the legacy dataset in `dataset.last_updated_at`

- This PR also gets the `geographic_coverage` value from legacy datasets and stores in in `location1`

### DIFF
--- a/app/helpers/manage_helper.rb
+++ b/app/helpers/manage_helper.rb
@@ -1,7 +1,7 @@
 module ManageHelper
 
   SORT_OPTIONS = {"-name" => {title: :desc}}
-  SORT_OPTIONS.default = {updated_at: :desc}
+  SORT_OPTIONS.default = {last_updated_at: :desc}
 
   def manage_sort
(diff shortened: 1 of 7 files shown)


hannako:
[Most recent filter not working - all datasets have last updated set to the same date - Aug 2017](https://trello.com/c/9u6o89zd/375-most-recent-filter-not-working-all-datasets-have-last-updated-set-to-the-same-date-aug-2017)

hannako:
[None of the datasets seem to displaying a geographic location field (all showing ‘Not available’) - is this a bug or bad data?](https://trello.com/c/nnZpMC3G/376-none-of-the-datasets-seem-to-displaying-a-geographic-location-field-all-showing-not-available-is-this-a-bug-or-bad-data)

PR #338, opened 2017-09-20
If on PaaS, get the ES server URL and the server's cert

maxf:
... from the VCAP_SERVICES environment variable

### DIFF
--- a/config/initializers/elasticsearch.rb
+++ b/config/initializers/elasticsearch.rb
@@ -1,9 +1,37 @@
-config = {
-  host: ENV.fetch("ES_HOST", "http://127.0.0.1:9200"),
-  transport_options: {
-    request: { timeout: 5 }
+require 'base64'
+
+vcap_services = ENV.fetch("VCAP_SERVICES", nil)
+
+if vcap_services
+  # we are running on PaaS
+  vcap = JSON.parse(vcap_services)
+
+  es_server = vcap['elasticsearch'][0]['credentials']['uri'].chomp("/")
+
+  es_cert = Base64.decode64(vcap['elasticsearch'][0]['credentials']['ca_certificate_base64'])
+  es_cert_file = File.new("/tmp/out.pem", "w")
+  es_cert_file.puts(es_cert)
+  es_cert_file.close
+
+  puts es_server
+  puts es_cert_file.path
+
+  config = {
+    host: es_server,
+    transport_options: {
+      request: { timeout: 5 },
+      ssl: { ca_file: es_cert_file.path }
+    }
   }
-}
+else
+  es_server = ENV.fetch("ES_HOST", "http://127.0.0.1:9200")
+  config = {
+    host: es_server,
+    transport_options: {
+      request: { timeout: 5 }
+    }
+  }
+end
 
 if File.exist?("config/elasticsearch.yml")
   config.merge!(YAML.load_file("config/elasticsearch.yml")[Rails.env].symbolize_keys)
--- a/config/initializers/elasticsearch.rb
+++ b/config/initializers/elasticsearch.rb
@@ -16,9 +16,8 @@ def create_es_cert_file(cert)
     es_cert_file.write(cert)
   ensure
     es_cert_file.close
-    es_cert_file.unlink
-    return es_cert_file
   end
+  es_cert_file
 end
 
 
--- a/config/initializers/elasticsearch.rb
+++ b/config/initializers/elasticsearch.rb
@@ -4,14 +4,13 @@ CONFIG_PATH = "#{Rails.root}/config/elasticsearch.yml".freeze
 TEMPLATE = ERB.new File.new(CONFIG_PATH).read
 ELASTIC_CONFIG = YAML.load(TEMPLATE.result(binding))[ENV['RAILS_ENV']]
 
-config = {
-  host: ELASTIC_CONFIG['host'],
-  transport_options: {
-    request: { timeout: 5 }
-  }
-}
+def log(server, filepath)
+  Rails.logger.info "Configuring Elasticsearch on PAAS.\n
+  Elasticsearch host: #{server}\n
+  Elasticsearch cert file path: #{filepath}"
+end
 
-if ELASTIC_CONFIG['vcap_services'].present?
+def es_config_production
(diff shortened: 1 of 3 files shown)


config/initializers/elasticsearch.rb
```diff
@@ -1,9 +1,37 @@
-config = {
-  host: ENV.fetch("ES_HOST", "http://127.0.0.1:9200"),
-  transport_options: {
-    request: { timeout: 5 }
+require 'base64'
+
+vcap_services = ENV.fetch("VCAP_SERVICES", nil)
```
laurentqro: `fetch` is useful if we want it to blow up if the key does not exist, or if we want to offer a fallback value. In this case, we could just write `ENV['VCAP_SERVICES']`, which will return `nil` if the `VCAP_SERVICES` key is not present in the `ENV` hash.

config/initializers/elasticsearch.rb
```diff
@@ -1,9 +1,37 @@
...
+
+if vcap_services
```
laurentqro: I'd suggest using the stricter `vcap_services.present?`. That protects us against the case when `vcap_services` is an empty string.

```
if ""
  "foo"
else
  "bar"
end

=> "foo"
```

```
"".present?
=> false
```

config/initializers/elasticsearch.rb
```diff
@@ -1,9 +1,37 @@
...
+  es_cert_file.puts(es_cert)
+  es_cert_file.close
+
+  puts es_server
+  puts es_cert_file.path
+
+  config = {
+    host: es_server,
+    transport_options: {
+      request: { timeout: 5 },
```
laurentqro: Should we put this `5` in an env variable?

config/initializers/elasticsearch.rb
```diff
@@ -1,9 +1,37 @@
...
+      ssl: { ca_file: es_cert_file.path }
+    }
   }
-}
+else
+  es_server = ENV.fetch("ES_HOST", "http://127.0.0.1:9200")
+  config = {
+    host: es_server,
+    transport_options: {
+      request: { timeout: 5 }
```
laurentqro: Not sure if that's important or not, but I have seen the docs write:

`Elasticsearch::Client.new request_timeout: 5`

https://github.com/elastic/elasticsearch-ruby/tree/master/elasticsearch-transport#setting-timeouts

TobyRet:
[Elasticsearch configuration for PAAS](https://trello.com/c/J1X4wHMG/501-elasticsearch-configuration-for-paas)

config/initializers/elasticsearch.rb
```diff
@@ -1,23 +1,53 @@
...
-  client = ::Dataset.__elasticsearch__.client
-  begin
-    client.indices.delete index: ::Dataset.__elasticsearch__.index_name
-  rescue
-    puts "No test search index to delete"
-  end
-  client.indices.create index: ::Dataset.__elasticsearch__.index_name
+def es_config_production
+  vcap = JSON.parse(ELASTIC_CONFIG['vcap_services'])
+
```
TobyRet: Boo. I get the point Code Climate but I think it's best to keep as it for readability

config/initializers/elasticsearch.rb
```diff
@@ -1,23 +1,53 @@
...
+  es_server = vcap['elasticsearch'][0]['credentials']['uri'].chomp('/')
+  es_cert = Base64.decode64(vcap['elasticsearch'][0]['credentials']['ca_certificate_base64'])
+  es_cert_file = File.new('/tmp/out.pem', 'w')
```
TobyRet: Max had some concerns about the naming of this file and potential duplicates / clashes on the server - @govuklaurence should it be randomised? Max also mentioned a Temp file class?

spec/spec_helper.rb
```diff
@@ -52,3 +61,66 @@
   end
 
 end
+
+def mappings
```
TobyRet: This mapping matches the one on find

config/initializers/elasticsearch.rb
```diff
@@ -1,23 +1,53 @@
...
+  es_cert_file.puts(es_cert)
+  es_cert_file.close
+
+  log(es_server, es_cert_file.path)
+
+  {
```
TobyRet: I'd prefer to take the hit for some duplication and write out each config individually rather than mutate a shared config

config/initializers/elasticsearch.rb
```diff
@@ -1,23 +1,53 @@
...
-  host: ENV.fetch("ES_HOST", "http://127.0.0.1:9200"),
-  transport_options: {
-    request: { timeout: 5 }
-  }
-}
+require 'base64'
+
+CONFIG_PATH = "#{Rails.root}/config/elasticsearch.yml".freeze
+TEMPLATE = ERB.new File.new(CONFIG_PATH).read
+ELASTIC_CONFIG = YAML.load(TEMPLATE.result(binding))[ENV['RAILS_ENV']]
```
laurentqro: How about just:

```
require 'base64'

CONFIG_PATH = Rails.root.join('config', 'elasticsearch.yml')
ELASTIC_CONFIG = YAML.load(File.read(CONFIG_PATH))[ENV['RAILS_ENV']]
```

config/initializers/elasticsearch.rb
govuklaurence: Yeah, use ruby's TempFile (https://docs.ruby-lang.org/en/2.4.0/Tempfile.html)

config/initializers/elasticsearch.rb
govuklaurence: 👍

config/initializers/elasticsearch.rb
```diff
@@ -1,23 +1,63 @@
...
   begin
-    client.indices.delete index: ::Dataset.__elasticsearch__.index_name
-  rescue
-    puts "No test search index to delete"
+    es_cert_file = Tempfile.new(%w(out .pem))
+    es_cert_file.write(cert)
+  ensure
+    es_cert_file.close
+    es_cert_file.unlink
+    return es_cert_file
```
govuklaurence: Return outside of block

PR #340, opened 2017-09-20
Delete the correct file

maxf:


laurentqro:
yes, afaik, the original `.gz` archive is still present after un-gzipping.

laurentqro:
Actually, just tested this out. No, the original `.gz` is gone after g-unzipping. So it's ok to just delete the `.jsonl` files.

PR #345, opened 2017-09-26
DRY up DatasetsController

laurentqro:
:point_up:

### DIFF
--- a/app/controllers/datasets_controller.rb
+++ b/app/controllers/datasets_controller.rb
@@ -90,8 +90,4 @@ class DatasetsController < ApplicationController
   def set_dataset
     @dataset = Dataset.find_by(:name => params.require(:id))
   end
-
-  def current_file
-    Datafile.find(params.require(:file_id))
-  end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/datasets_controller.rb
```diff
@@ -1,8 +1,9 @@
 class DatasetsController < ApplicationController
   before_action :authenticate_user!
+  before_action :set_dataset, only: [:show, :edit, :update, :destroy,
```
govuklaurence: That's a lot of onlies :) you can use `except` instead to exclude things

laurentqro: Yeah not great indeed, but I'd rather maintain a whitelist than a blacklist; I don't want the callback to automatically apply to any new controller actions.

I was also considering removing the callback altogether and having a `set_dataset` call in all the controller actions that need it.

govuklaurence: That makes sense 👍

PR #349, opened 2017-09-29
Fix sorting on manage datasets page

laurentqro:
Main change here is to replace the `published` boolean field on the `datasets` table with a `status` enum field, with two statuses: `draft` and `published`.

This gives us these kinds of niceties for free, out of the box:

```
dataset.published!
dataset.published? # => true
dataset.status  # => "published"
```

http://api.rubyonrails.org/v5.1/classes/ActiveRecord/Enum.html

And it makes sorting datasets by `status` on the `/manage` page very easy. It removes the need to hack a sorting by status functionality using grouping (https://github.com/datagovuk/publish_data_beta/pull/349/files#diff-615c7fe6f11b982b52d08a41049a6909L20).

This PR also has a rake task to migrate the data from the `published` boolean field to the new `status` enum field.

We'll probably want to do the same refactoring with Dataset `stage`.


db/migrate/20171004101315_remove_published_column_from_datasets.rb
```diff
@@ -1,5 +0,0 @@
-class RemovePublishedColumnFromDatasets < ActiveRecord::Migration[5.1]
-  def change
-    remove_column :datasets, :published
```
laurentqro: we'll do this in a separate PR, once the data migration is done.

lib/tasks/datasets.rake
```diff
@@ -0,0 +1,25 @@
+namespace :datasets do
```
laurentqro: this task can be deleted once it has run in production.

app/controllers/manage_controller.rb
```diff
@@ -1,20 +1,19 @@
...
 
     if @q.present?
-      @datasets = @datasets.where(sql_query, query: @q.downcase).page(params[:page])
-    else
-      @datasets.page(params[:page])
+      @datasets = @datasets.where(sql_query, query: @q.downcase)
     end
 
-    manage_sort
+    @datasets = @datasets.order("#{sort_column} #{sort_direction}").page(params[:page])
```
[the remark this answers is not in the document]
laurentqro: I am whitelisting the parameters to just `asc` and `desc` for sort direction and to the table column names for the sort by column, so I think we're covered.

PR #350, opened 2017-09-29
Vcap config fix

maxf:
I think it's more flexible to allow the app to read from VCAP_SERVICES even locally. That way it's possible to access a PaaS elasticsearch server by setting VCAP_SERVICES in your terminal.

Other changes are around better error handling. See respective commits.

### DIFF
--- a/config/initializers/elasticsearch.rb
+++ b/config/initializers/elasticsearch.rb
@@ -25,7 +25,7 @@ def create_es_cert_file(cert)
 end
 
 
-def es_config_production
+def es_config_from_vcap
   begin
     vcap = JSON.parse(ELASTIC_CONFIG['vcap_services'])
   rescue => e
@@ -67,7 +67,7 @@ def es_config_production
   }
 end
 
-def es_config_non_production
+def es_config_from_host
   {
     host: ELASTIC_CONFIG['host'],
     transport_options: {
@@ -79,8 +79,8 @@ def es_config_non_production
 end
 
 config = ELASTIC_CONFIG['vcap_services'].present? ?
-            es_config_production :
-            es_config_non_production
+            es_config_from_vcap :
+            es_config_from_host
 
 ELASTIC = Elasticsearch::Client.new(config)
 Elasticsearch::Model.client = ELASTIC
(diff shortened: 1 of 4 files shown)


config/initializers/elasticsearch.rb
```diff
@@ -1,9 +1,15 @@
 require 'base64'
-require 'tempfile'
 
 CONFIG_PATH = Rails.root.join('config', 'elasticsearch.yml')
 TEMPLATE = ERB.new File.new(CONFIG_PATH).read
-ELASTIC_CONFIG = YAML.load(TEMPLATE.result(binding))[ENV['RAILS_ENV']]
+
+begin
```
TobyRet: nice improvement

config/initializers/elasticsearch.rb
```diff
@@ -13,22 +19,29 @@ def log(server, filepath)
 
 def create_es_cert_file(cert)
   begin
-    es_cert_file = Tempfile.new(%w(out .pem))
+    es_cert_file = File.new('elasticsearch_cert.pem', 'w')
```
TobyRet: Why is the Tempfile class not being used anymore?

config/initializers/elasticsearch.rb
```diff
@@ -56,8 +69,8 @@ def es_config_non_production
 end
 
 config = ELASTIC_CONFIG['vcap_services'].present? ?
-            es_config_production :
-            es_config_non_production
+            es_config_from_vcap :
```
TobyRet: It's now not clear if you are connection to production elasticsearch or locally

config/initializers/elasticsearch.rb
maxf: Because the file gets deleted randomly. It's not too clear from the documentation when they're deleted, but we shouldn't assume they remain after we've closed them.

config/initializers/elasticsearch.rb
maxf: That's the question I'm asking in the original comment above. Why not support VCAP_SERVICES locally? This would be useful when you want to populate a PaaS ES index from your local machine. We need to discuss generally about Rails's environments and our own environments: we may run in production mode on our staging server, etc.

maxf: Moreover, I don't think there should be code that's environment specific anywhere in the application. It opens the door to things behaving differently depending on what environment it's in. Code coverage and testing also become difficult as you have tests that also depend on the environment.

PR #357, opened 2017-10-09
Dry up specs with factories + fix skipped failing test

laurentqro:
☝️

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -8,7 +8,7 @@ class Link < Datafile
   has_one :preview, foreign_key: "datafiles_id", dependent: :destroy
 
   before_save :set_dates
-  after_commit :generate_preview_async
+  after_save :generate_preview_async
 
   validates :quarter, presence: true,
     if: -> { !self.documentation && self.dataset.quarterly? }
@@ -35,6 +35,7 @@ class Link < Datafile
   end
 
   private
+
   def generate_preview_async
     PreviewGenerationWorker.perform_async(self.id)
   end
--- a/spec/features/create_dataset_spec.rb
+++ b/spec/features/create_dataset_spec.rb
@@ -22,7 +22,7 @@ describe "dataset creation" do
     end
 
     it "publishes a dataset" do
-      visit "/datasets/new"
+      visit new_dataset_path
 
       # PAGE 1: New
       fill_in "dataset[title]", with: "my test dataset"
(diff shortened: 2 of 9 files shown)


app/models/link.rb
```diff
@@ -8,7 +8,7 @@ class Link < Datafile
   has_one :preview, foreign_key: "datafiles_id", dependent: :destroy
 
   before_save :set_dates
-  after_commit :generate_preview_async
```
laurentqro: This is what was making the (skipped) test fail. We were trying to generate a preview after every DB commit, including when a link had been deleted. So when we are trying to generate a preview for a non-existing link, we get an `record not found` error.

spec/features/create_dataset_spec.rb
```diff
@@ -22,7 +22,7 @@
     end
 
     it "publishes a dataset" do
-      visit "/datasets/new"
```
laurentqro: removing hard-coded URLs means those could be modified in the future without having to update the tests each time.

spec/features/edit_dataset_spec.rb
```diff
@@ -21,7 +21,7 @@
   context 'editing published datasets from show page' do
     before(:each) do
       click_link 'Manage datasets'
-      edit_dataset(:published_dataset)
```
laurentqro: this `edit_dataset` helper was dependent on order, which made it brittle. Fetching the link by its `href` instead.

spec/features/edit_dataset_spec.rb
```diff
@@ -81,21 +81,23 @@
...
       click_button 'Save and continue'
-
       expect(page).to have_content('my other test file')
     end
 
     it "should be able to edit an existing file" do
+      link = published_dataset.links.first
+
       click_change(:datalinks)
-      expect(page).to have_content("my published test file")
```
laurentqro: I removed all hardcoded titles/names so we can have more flexibility to change our factories without breaking the tests.

spec/support/dataset_helper.rb
```diff
@@ -25,44 +15,30 @@ def click_change(property)
 
 def set_up_models
 
-    let(:land) { FactoryGirl.create(:organisation, name: 'land-registry', title: 'Land Registry') }
+    let(:land) { FactoryGirl.create(:organisation) }
     let(:user) { FactoryGirl.create(:user, primary_organisation: land) }
+
     let(:published_dataset) { FactoryGirl.create(:dataset,
```
laurentqro: making FactoryGirl work harder for us :)

PR #358, opened 2017-10-09
Create endpoint for syncing legacy datasets

TobyRet:
This is an updated PR. In response to feedback, api keys and associated model has been removed.  Chron job will need to add correct username and password as parameter urls as basic authentication is set for staging and production environments

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -2,6 +2,10 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
 
+  if Rails.env.production? || Rails.env.staging?
+    http_basic_authenticate_with name: ENV["HTTP_USERNAME"], password: ENV["HTTP_PASSWORD"]
+  end
+
   private
 
   def record_not_found
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,26 +1,10 @@
-require 'rake'
-
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
-
-  if [Rails.env.production?, ENV["HTTP_USERNAME"], ENV["HTTP_PASSWORD"]].all?
-    http_basic_authenticate_with name: ENV["HTTP_USERNAME"], password: ENV["HTTP_PASSWORD"]
-  end
 
   private
 
   def record_not_found
     render plain: '404 Not Found', status: 404
-  end
-
-  def call_rake(task, options = {})
-    PublishDataBeta::Application.load_tasks
-
-    log_path = %Q(#{Rails.root}/log/rake.log)
-    options[:rails_env] ||= Rails.env
-    args = options.map { |n, v| %Q(#{n.to_s.upcase}='#{v}') }
-
-    system %Q(rake #{task} #{args.join(' ')} --trace 2>&1 >> #{log_path} &)
   end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,13 +1,26 @@
+require 'rake'
+
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   rescue_from ActiveRecord::RecordNotFound, with: :record_not_found
 
   if [Rails.env.production?, ENV["HTTP_USERNAME"], ENV["HTTP_PASSWORD"]].all?
-      http_basic_authenticate_with name: ENV["HTTP_USERNAME"], password: ENV["HTTP_PASSWORD"]
+    http_basic_authenticate_with name: ENV["HTTP_USERNAME"], password: ENV["HTTP_PASSWORD"]
   end
 
   private
+
   def record_not_found
-    render plain: "404 Not Found", status: 404
+    render plain: '404 Not Found', status: 404
+  end
+
+  def call_rake(task, options = {})
+    PublishDataBeta::Application.load_tasks
+
+    log_path = %Q(#{Rails.root}/log/rake.log)
+    options[:rails_env] ||= Rails.env
+    args = options.map { |n, v| %Q(#{n.to_s.upcase}='#{v}') }
+
+    system %Q(rake #{task} #{args.join(' ')} --trace 2>&1 >> #{log_path} &)
   end
 end
--- a/app/controllers/sync_controller.rb
+++ b/app/controllers/sync_controller.rb
@@ -1,0 +1,7 @@
+class SyncController < ApplicationController
+  def legacy
+    call_rake('sync:daily')
+    msg = { status: 200, body: 'Syncing legacy datasets' }
+    render json: msg
+  end
+end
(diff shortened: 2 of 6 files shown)


maxf:
@TobyRet  why do we need API keys in the model?

lib/tasks/sync.rake
```diff
@@ -10,11 +10,15 @@ namespace :sync do
...
 
+    puts "#{Time.now} - START: Attempting to sync legacy datasets"
+
     get_packages "https://data.gov.uk" do |package|
       MetadataTools.add_dataset_metadata(package, orgs_cache, theme_cache)
-      print "Imported #{count+=1} datasets...\r"
+      puts "Imported #{count+=1} datasets...\r"
     end
 
+    puts "#{Time.now} - SUCCESS: Finished syncing #{count} legacy datasets"
```
laurentqro: maybe worth writing this to the logs?

```
  logger = Logger.new
  logger.info("#{Time.now} - START: Attempting to sync legacy 
  datasets")

  # rake task runs ...

  logger.info("#{Time.now} - SUCCESS: Finished syncing #{count} 
  legacy datasets")
```

laurentqro: also a good idea to log any errors that we might get during the sync rake task.

app/controllers/application_controller.rb
```diff
@@ -10,4 +10,10 @@ class ApplicationController < ActionController::Base
   def record_not_found
     render plain: "404 Not Found", status: 404
   end
+
+  def call_rake(task, options = {})
+    options[:rails_env] ||= Rails.env
+    args = options.map { |n, v| "#{n.to_s.upcase}='#{v}'" }
```
govuklaurence: Make sure to escape any args

app/controllers/sync_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+  def legacy
+    call_rake('sync:daily')
+    msg = { status: 200, body: 'Syncing legacy datasets' }
+    render json: msg
+  end
+
+  private
+
+  def restrict_access
+    authenticate_or_request_with_http_token do |token, options|
```
govuklaurence: Remove unused arg

app/controllers/application_controller.rb
```diff
@@ -10,4 +10,10 @@ class ApplicationController < ActionController::Base
...
+    system "rake #{task} #{args.join(' ')} --trace 2>&1 >> #{Rails.root}/log/rake.log &"
```
govuklaurence: I think Rake has a facility to call tasks within its lib -- avoid system calls if possible

app/models/api_key.rb
```diff
@@ -0,0 +1,11 @@
+class ApiKey < ApplicationRecord
+  before_create :generate_access_token
+
+  private
+
+  def generate_access_token
+    begin
+      self.access_token = SecureRandom.hex
+    end while self.class.exists?(access_token: access_token)
```
govuklaurence: Always prefer `loop` over `while`

lib/tasks/sync.rake
```diff
@@ -10,11 +10,15 @@ namespace :sync do
     theme_cache = Theme.all.pluck(:title, :id).to_h
     count = 0
 
+    puts "#{Time.now} - START: Attempting to sync legacy datasets"
```
govuklaurence: Agree with @laurentqro -- use rails logger

app/models/api_key.rb
```diff
@@ -0,0 +1,11 @@
+class ApiKey < ApplicationRecord
+  before_create :generate_access_token
+
+  private
+
+  def generate_access_token
+    begin
+      self.access_token = SecureRandom.hex
```
laurentqro: `hex` only allows 0-9 and a-f. You'll have a wider character set with Base64: `SecureRandom.urlsafe_base64(length)`

app/controllers/application_controller.rb
```diff
@@ -10,4 +10,10 @@ class ApplicationController < ActionController::Base
...
+  end
```
laurentqro: How about something like:

def call_rake(task_name)
  Rake::Task[task_name].invoke
  logger = Logger.new('legacy_sync.log')
  logger.info "#{Time.now} -- #{name}"
end

app/models/api_key.rb
laurentqro: 👍

app/controllers/sync_controller.rb
```diff
@@ -0,0 +1,19 @@
+require 'legacy_data_sync'
+
+class SyncController < ApplicationController
```
laurentqro: nitpicking here, but maybe namespace all our API stuff?

e.g. `API::Datasets::SynchronizationController`, under `app/controllers/api/datasets/synchronization_controller.rb`

govuklaurence: Overkill, especially since this is a temporary endpoint

config/routes.rb
```diff
@@ -16,6 +16,8 @@
   get 'tasks', to: 'tasks#my'
   get 'tasks/organisation', to: 'tasks#organisation'
 
+  get 'api/legacy_sync', to: 'sync#legacy'
```
laurentqro: still in relation to namespacing API stuff, maybe consider:

```
namespace :api do
  ...
end
```

app/models/api_key.rb
laurentqro: why? just curious :)

maxf: Codeclimate says "Use `Kernel#loop` with `break` rather than `begin/end/until`(or `while`)."

lib/tasks/sync.rake
TobyRet: Cool

app/controllers/application_controller.rb
TobyRet: @govuklaurence - the reason why as system call is being used is so that the process can be run in the background. I have not been able to find a way to do this with Rake::Task - any ideas?

app/models/api_key.rb
TobyRet: No longer required  - using basic auth

app/controllers/application_controller.rb
TobyRet: @laurentqro - see my comment above to @govuklaurence re: system calls

app/controllers/sync_controller.rb
TobyRet: No longer required

config/routes.rb
TobyRet: Same

lib/tasks/sync.rake
```diff
@@ -10,11 +10,19 @@ namespace :sync do
     theme_cache = Theme.all.pluck(:title, :id).to_h
     count = 0
 
+    puts "#{Time.now} - Starting legacy data sync"
```
TobyRet: puts / print statements are being piped to log file (set in SyncController)

config/routes.rb
```diff
@@ -16,6 +16,8 @@
...
+
```
maxf: It's only temporary, so I won't be too pedantic about it, but it's often a good idea to name API calls with a verb if it's an action with changes (and make it a POST too), eg 'start_sync' or something.

However could you add a comment about what this does and the fact it's temporary?

TobyRet: Yes of course - will change this

PR #359, opened 2017-10-09
Harvested datasets are readonly

laurentqro:
Each time a dataset is harvested, it is updated in Publish. If a publisher edits a harvested dataset, the changes will be overwritten at the next harvest. To prevent that, we want to make such datasets readonly.

This PR

* checks if dataset is harvested
* hides Add Data and Edit links if dataset is harvested
* aborts update of a harvested dataset and displays error message to the user

Note that the import and sync rake tasks skip validation when updating a dataset, so we're all good on that front.

https://trello.com/c/Fclrjhei/5-harvested-datasets-are-read-only-on-publish-beta

screen shot 2017-10-11 at 08 39 00

screen shot 2017-10-11 at 11 06 10

This PR prevents harvested datasets from being edited by the user from the UI - i.e. the 'edit' button is hidden in the view and the guard clause is present only in the `update` controller action.

However, a harvested dataset can still be updated programmatically at the model layer level, e.g. with: `harvested_dataset.update(title: "new title")`.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -34,10 +34,18 @@ class Dataset < ApplicationRecord
   validates :licence, presence: true, if: :published?
   validates :licence_other, presence: true, if: lambda { licence == 'other' }
   validates :stage, inclusion: { in: STAGES }
-  validate :published_dataset_must_have_datafiles_validation
+
+  validate  :published_dataset_must_have_datafiles_validation
+  validate  :is_readonly?, on: :update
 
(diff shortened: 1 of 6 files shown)


govuklaurence:
Looking good, but I'd like to see one more test that tries to write/update the dataset and errors

govuklaurence:
Hmm. As long as it still allows us to pull in sync updates.

govuklaurence:
Looks good. Sync should still work as it currently ignores validations 👍  good to go

PR #369, opened 2017-10-12
Updates to metadata on publish sent to legacy

hannako:
This PR is part of [this trello card](https://trello.com/c/7XtWXb4p/3-sync-changes-to-existing-datasets-on-publish-to-legacy)

There are three requirements:

- updates to dataset metadata are sent to legacy 
- updates to datafiles are sent to legacy
- creation of new datasets and datafiles are sent to legacy

This PR implements the first requirement. Comments v welcome on how best to test this @govuklaurence ! 

Have also made the users created by seeds.rb belong to real organisations so that their datasets populate the manage_datasets page

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -17,6 +17,7 @@ class Dataset < ApplicationRecord
   after_initialize :set_initial_stage
   before_destroy :prevent_if_published
   before_save :set_uuid
+  after_save :update_legacy
 
   belongs_to :organisation
   belongs_to :theme, optional: true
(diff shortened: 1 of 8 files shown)


app/workers/legacy_sync_worker.rb
```diff
@@ -0,0 +1,15 @@
+require 'rest-client'
+
+class LegacySyncWorker
+  include Sidekiq::Worker
+
+  def perform(id)
+    json = Dataset.find(id).ckanify_metadata.to_json
+    server =
+      Rails.env.production? ? "https://data.gov.uk" : "https://test.data.gov.uk"
```
hannako: The worker is only called if the env is production, but have added as a second line of defence as I was nervous of breaking legacy! Perhaps there's a better safety mechanism...

app/models/dataset.rb
```diff
@@ -78,6 +78,46 @@ def as_indexed_json(_options={})
     )
   end
 
+  # map the dataset.as_json so that the keys are in a format that ckan recognises
+  def ckanify_metadata
```
TobyRet: Are we 100% sure all those fields will be present?

spec/models/dataset_spec.rb
```diff
@@ -95,4 +95,35 @@
 
     expect(Dataset.count).to eq 0
   end
+
+  it "can ckanify its metadata" do
```
TobyRet: Related to the comment above. This tests the 'happy' path but do we need to think of edge cases?

spec/workers/legacy_sync_worker.rb
```diff
@@ -0,0 +1,6 @@
+require 'rails_helper'
+
```
TobyRet: I see there is a testing library for sidekiq. https://github.com/mperham/sidekiq/wiki/Testing
Would be good to hear from the others whether they feel its worth adding tests for all workers.

app/workers/legacy_sync_worker.rb
maxf: This means running tests requires test.data.gov.uk? That's probably not ideal as it depends on test.dgu being up, but also we cannot run the tests offline. Is mocking a possibility?

laurentqro: yeah I think it might be a good idea to mock data.gov.uk in our tests here.

spec/workers/legacy_sync_worker.rb
govuklaurence: It would be helpful, but as long as the lib code works it should be good to go. Workers should really only ever call code inside the app

db/seeds.rb
```diff
@@ -33,15 +33,9 @@
...
-land_registry.title = "Land Registry"
-land_registry.save!()
-
-hmrc = Organisation.new
-hmrc.name = "hmrc"
-hmrc.title = "hmrc"
-hmrc.save!()
+land_registry = Organisation.find_by(name: 'land-registry')
+hmrc = Organisation.find_by(name: 'hmrc')
+gds = Organisation.find_by(name: 'government-digital-service')
```
laurentqro: This will not find anything if it's the first time I run rake db:seed.

spec/workers/legacy_sync_worker.rb
laurentqro: yeah, I'd test the logic that the worker runs, rather than test the worker itself, if that makes sense.

app/workers/legacy_sync_worker.rb
```diff
@@ -0,0 +1,15 @@
...
+    url = "#{ server }/api/3/action/package_patch"
+    headers = {Authorization: "6c25593f-19fe-4064-a451-ddc0aff562d9"}
```
govuklaurence: this needs to be in an env var, since it's a secret

setup.sh
```diff
@@ -14,7 +14,7 @@ print_error() {
 
 export SECRET_KEY_BASE="foobar"
 bundle install
-rails db:reset db:migrate db:seed
+rails db:drop db:create db:schema:load
```
govuklaurence: still needs to `db:seed`

hannako: 👍

PR #372, opened 2017-10-16
Refactoring suggestion

laurentqro:


app/models/legacy/server.rb
```diff
@@ -0,0 +1,23 @@
+class Legacy::Server
+  def update(payload)
+    RestClient.put(url, payload, headers)
```
govuklaurence: Log if the PUT fails

app/models/legacy/dataset.rb
```diff
@@ -0,0 +1,46 @@
+class Legacy::Dataset < SimpleDelegator
+  def json_metadata
+    {
+      'id': uuid,
```
govuklaurence: this line will have a symbol key

app/models/legacy/server.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  private
+
+  def url
+    "https://#{host}/#{path}"
```
govuklaurence: Use `URI::join`

hannako:
Closing this PR so I can delete branch. I will implement this refactor on my branch and open a fresh PR

PR #374, opened 2017-10-17
Refactor sync:daily rake task

TobyRet:
Logic extracted to its own class to make it easier to test and add new functionality.

Next PR will be to add logic for indexing datasets from legacy in Elasticsearch ...

### DIFF
--- a/lib/tasks/sync.rake
+++ b/lib/tasks/sync.rake
@@ -1,52 +1,19 @@
 require 'rest-client'
-require 'util/metadata_tools'
+require './lib/util/legacy_dataset_sync'
 
 namespace :sync do
-
-  desc "Synchronise datasets from Legacy DGU"
-  task :daily => :environment do |_, args|
-
+  desc 'Synchronise datasets from Legacy DGU'
+  task legacy: :environment do |_, args|
     orgs_cache =  Organisation.all.pluck(:uuid, :id).to_h
     theme_cache = Theme.all.pluck(:title, :id).to_h
-    count = 0
+    host = Rails.env.production? ? 'https://data.gov.uk' : 'https://test.data.gov.uk'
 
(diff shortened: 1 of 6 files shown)


lib/tasks/sync.rake
```diff
@@ -1,52 +1,19 @@
 require 'rest-client'
-require 'util/metadata_tools'
+require './lib/util/legacy_dataset_sync'
```
govuklaurence: avoid relative imports -- `util/legacy_dataset_sync` should work

spec/lib/util/legacy_dataset_sync_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+      theme_cache: theme_cache,
+      host: host
+    )
+
+    stub_request(:get, url).to_return(status: 200, body: response)
+    allow(MetadataTools).to receive(:add_dataset_metadata)
+
+    legacy_dataset_sync.run
+
+    WebMock
```
govuklaurence: Don't expectations work with webmock?

spec/lib/util/legacy_dataset_sync_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+require './lib/util/legacy_dataset_sync'
+
+describe LegacyDatasetSync do
+  it 'Imports legacy datasets' do
+    orgs_cache =  Organisation.all.pluck(:uuid, :id).to_h
+    theme_cache = Theme.all.pluck(:title, :id).to_h
+
+    host = 'www.legacyServer.com'
+    path = '/api/3/action/package_search?q=metadata_modified:%5BNOW-1DAY%20TO%20NOW%5D&rows=5000'
+    url = "http://#{host}#{path}"
```
govuklaurence: Use `URI::join`

PR #376, opened 2017-10-17
Jess/datasync publish to legacy

maxf:
- metadata updates made on publish to existing legacy datasets are sent to ckan
- datafile(resource) updates will be in separate PR


app/models/dataset.rb
```diff
@@ -78,6 +78,46 @@ def as_indexed_json(_options={})
...
+    publish_json = self.as_json
+    organisation = Organisation.find(publish_json["organisation_id"])
+    ckan_json = {
+      id: publish_json['uuid'],
+      name: publish_json['name'],
+      title: publish_json['title'],
+      notes: publish_json['summary'],
+      description: publish_json['summary'],
+      organization: {name: organisation.name},
+      update_frequency: convert_freq_to_legacy_format(publish_json['frequency']),
```
rossjones: This needs to be sent as an 'extra' (a key value that is used to extend the schema).  Although extras are promoted to the top level in package_show the same isn't true in reverse...

```
extras: [{key: "update_frequency", value: "...." }]
```

You may find others (like unpublished) may be the same but unpublished in particular is probably unnecessary for new system.

hannako: Great, thanks Ross, that makes sense.

app/models/legacy/dataset.rb
```diff
@@ -0,0 +1,67 @@
...
+        "package_id" => uuid,
+        "value" => ckan_dataset["update_frequency-other"]
+      }
+      ckan_dataset["extras"].push(extra)
+    end
+    ckan_dataset
+  end
+
+  def unpublished?(string)
+    string == "draft"
```
laurentqro: @hannako the day we add another status (like "discarded"), this will return `true`. The only way to tell whether a dataset is unpublished is if its status is anything else than `"published"`.

PR #379, opened 2017-10-18
Treat links and docs as resources under datasets

laurentqro:
First step in a larger cleanup of the routes file.


config/routes.rb
```diff
@@ -19,21 +19,15 @@
   get 'api/start_legacy_sync', to: 'sync#legacy'
 
   resources :datasets do
-    get 'show/:id', to: 'datasets#show'
```
laurentqro: obsolete route

PR #388, opened 2017-10-23
Keep track of slug history to redirect to latest URL

laurentqro:
As it is, changing the title of a dataset modifies the dataset's slug. Which means we break any external links pointing to the old URL.

This PR introduces the `history` functionality of the FriendlyId gem. It keeps track of the previous versions of slugs in a dedicated table, which allows us to redirect any broken links to the latest version of the URL.

--

https://trello.com/c/bTQ1Nspf/48-in-publish-beta-redirect-dataset-urls-with-old-slugs-to-the-latest-version-the-url-in-order-to-avoid-broken-links-when-a-dataset

http://norman.github.io/friendly_id/file.Guide.html#History__Avoiding_404_s_When_Slugs_Change

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -16,7 +16,7 @@ class Dataset < ApplicationRecord
   include Elasticsearch::Model
   extend FriendlyId
 
-  friendly_id :uuid_and_title, :use => :slugged, :slug_column => :name
+  friendly_id :uuid_and_title, use: :history, slug_column: :name
   index_name ENV['ES_INDEX'] || "datasets-#{Rails.env}"
   document_type "dataset"
 
--- a/config/initializers/friendly_id.rb
+++ b/config/initializers/friendly_id.rb
@@ -1,0 +1,92 @@
+# FriendlyId Global Configuration
+#
+# Use this to set up shared configuration options for your entire application.
+# Any of the configuration options shown here can also be applied to single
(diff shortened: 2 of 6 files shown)


app/models/dataset.rb
```diff
@@ -16,7 +16,7 @@ def self.columns
   include Elasticsearch::Model
   extend FriendlyId
 
-  friendly_id :uuid_and_title, :use => :slugged, :slug_column => :name
+  friendly_id :uuid_and_title, use: :history, slug_column: :name
```
laurentqro: http://norman.github.io/friendly_id/file.Guide.html#History__Avoiding_404_s_When_Slugs_Change

app/models/dataset.rb
```diff
@@ -105,7 +105,7 @@ def uuid_and_title
   end
 
   def should_generate_new_friendly_id?
-    title_changed?
+    name.blank?
```
laurentqro: We generate a new slug only if the `name` field (where we store slugs) is blank. That way, the slugs from legacy stay the same in Beta, and we avoid overwriting the slugs in Legacy when we sync datasets from Beta => Legacy.

The cost of this (for now): changing the title of a dataset in Beta does not generate a new slug.

laurentqro:
closing this. replaced with https://github.com/datagovuk/publish_data_beta/pull/391.

PR #390, opened 2017-10-24
Store legacy dataset name on publish beta dataset

laurentqro:
This makes sure we can:

* redirect any links to legacy to its new URL in the new Publish app
* avoid overwriting the names (slugs) in legacy, which breaks links

### DIFF
--- a/app/models/legacy/dataset.rb
+++ b/app/models/legacy/dataset.rb
@@ -1,10 +1,9 @@
 class Legacy::Dataset < SimpleDelegator
 
   def metadata_json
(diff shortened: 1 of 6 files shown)


app/models/legacy/dataset.rb
```diff
@@ -1,10 +1,9 @@
 class Legacy::Dataset < SimpleDelegator
 
   def metadata_json
-    organisation = Organisation.find(organisation_id)
```
laurentqro: not necessary with `SimpleDelegator`

PR #391, opened 2017-10-25
Homemade slugs with custom routes

laurentqro:
This PR introduces:

* new URLs for datasets. The URL structure is now: `/datasets/:uuid/:name`, where name acts as the datasets URL slug,

* generates a new slug every time a dataset's title changes,

* redirects of any dataset URL with an old slug to the URL with the latest up to date slug. The slug is now entirely cosmetic. As long as the `:uuid` is correct in `/datasets/:uuid/:name`, the user will be directed to the correct dataset.

trello: https://trello.com/c/bTQ1Nspf/48-in-publish-beta-redirect-dataset-urls-with-old-slugs-to-the-latest-version-the-url-in-order-to-avoid-broken-links-when-a-dataset

### DIFF
--- a/app/controllers/datasets/docs_controller.rb
+++ b/app/controllers/datasets/docs_controller.rb
@@ -14,7 +14,7 @@ class Datasets::DocsController < ApplicationController
     @doc = @dataset.docs.build(doc_params)
 
     if @doc.save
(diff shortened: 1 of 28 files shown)


govuklaurence:
Or `scope` perhaps

PR #395, opened 2017-10-27
Post datafile updates to legacy

laurentqro:
* send dataset modifications to legacy every time it is published.
* sync all datafiles that were modified since the last publication of their parent dataset

CKAN docs: http://docs.ckan.org/en/latest/api/index.html#ckan.logic.action.patch.resource_patch.

### To do

* [x] Merge https://github.com/datagovuk/publish_data_beta/pull/402 to master and rebase
--

Trello: https://trello.com/c/7XtWXb4p/3-sync-changes-to-existing-datasets-on-publish-to-legacy


spec/rails_helper.rb
```diff
@@ -1,6 +1,8 @@
 # This file is copied to spec/ when you run 'rails generate rspec:install'
 require 'spec_helper'
 ENV['RAILS_ENV'] ||= 'test'
+ENV['LEGACY_HOST'] ||= "https://test.data.gov.uk"
+ENV['LEGACY_API_KEY'] ||= "123"
```
laurentqro: is this still necessary if we use a `.env.test` file with the dotenv gem?

app/models/dataset.rb
```diff
@@ -60,11 +60,27 @@ def publish!
       transaction do
         self.published!
         PublishingWorker.perform_async(self.id)
-        self.published_date = Time.now
+        self.last_published_at = Time.now
       end
     end
   end
 
+def update_legacy
```
laurentqro: May be good to move those syncing methods to a service object. What do you think?

spec/rails_helper.rb
govuklaurence: CI doesn't have dotenv

spec/models/dataset_spec.rb
```diff
@@ -103,4 +103,29 @@
     expect(dataset.last_published_at).to eq second_publish
   end
 
+  it "only updates legacy with datafiles that have been updated" do
```
laurentqro: maybe something more detailed, like "only updates legacy with datafiles that have been updated since latest dataset publication" or something along these lines.

spec/rails_helper.rb
laurentqro: :(

govuklaurence: Add it if you want to -- or check, because it might, but it wasn't built in mind with supporting dotenv

laurentqro: might add support for it in another PR

govuklaurence:
@laurentqro is this good to go?

app/models/legacy/server.rb
```diff
@@ -1,29 +1,17 @@
...
-    URI.join(host, path).to_s
-  end
-
   def host
-    ENV['LEGACY_HOST']
-  end
-
-  def path
-    "/api/3/action/package_patch"
+    ENV.fetch('LEGACY_HOST')
```
govuklaurence: blank fetch will just return the same, surely?

app/models/legacy/server.rb
```diff
@@ -1,29 +1,17 @@
...
   end
 
   def headers
-    { Authorization: ENV['LEGACY_API_KEY'] }
+    { Authorization: ENV.fetch('LEGACY_API_KEY') }
```
govuklaurence: Same here

app/services/legacy_sync_service.rb
```diff
@@ -0,0 +1,31 @@
+class LegacySyncService
```
govuklaurence: this class may better be expressed as a module... it's got one instance var and otherwise just runs methods of other objects

app/workers/legacy_update_worker.rb
```diff
@@ -0,0 +1,12 @@
+class LegacyUpdateWorker
+  include Sidekiq::Worker
+
+  def perform(url, payload, headers)
+    begin
+      RestClient.post(url, payload, headers)
+    rescue => e
+      Rails.logger.debug "Failed to send update request to Legacy with error: #{e.message}"
```
govuklaurence: `Rails.logger.error`

app/services/legacy_sync_service.rb
hannako: I thought the reason for having a class over a module was because class stores state, and module doesn't. But we are initialising with a dataset, so does that not suggest class is better? Am just asking for learning purposes!

app/models/legacy/server.rb
laurentqro: I want this to fail if the `LEGACY_HOST` key is absent. In order to avoid passing a nil` all the way down to `RestCient#post`.

app/workers/legacy_update_worker.rb
laurentqro: yep good idea

app/services/legacy_sync_service.rb
govuklaurence: My argument is that the state doesn't need to be stored, and everywhere this class is being used it's being called like a function (initialise followed by immediate method call)

app/models/legacy/server.rb
govuklaurence: Ah, cool.

app/services/legacy_sync_service.rb
laurentqro: I agree, it doesn't need to be stored, but it's nice to :)

laurentqro: I prefer to use a module as a "bag of methods" that I can include into multiple classes to extend them.

In this case, class or module is just a matter of preference.

PR #398, opened 2017-11-01
Fix legacy to beta sync (for new and modified datasets)

TobyRet:


lib/util/metadata_tools.rb
```diff
@@ -45,6 +45,7 @@ def add_dataset_metadata(obj, orgs_cache, theme_cache)
 
     d.status = "published"
     d.save!(validate: false)
+    d.id
```
laurentqro: where will be used?

TobyRet: For the elasticsearch publishing worker

PR #402, opened 2017-11-03
Remove dataset weekly frequency + add dates to daily datafiles

laurentqro:
There is no clear user need (legacy only has 10 datasets marked as weekly). It makes full sync with legacy possible, resulting in an easier transition for publishers.

This PR removes the `weekly` option for datasets and introduces the requirement for dates when creating datafiles for daily datasets.

Trello: https://trello.com/c/DTTDnTUv/53-remove-weekly-frequency-from-datasets

### DIFF
--- a/app/controllers/datasets/links_controller.rb
+++ b/app/controllers/datasets/links_controller.rb
@@ -60,8 +60,7 @@ class Datasets::LinksController < ApplicationController
     params.require(:link).permit(
       :url,
       :name,
-      :start_day, :start_month, :start_year,
-      :end_day, :end_month, :end_year,
+      :day, :month, :year,
       :year,
       :quarter
     )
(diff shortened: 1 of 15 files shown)


app/controllers/datasets/links_controller.rb
```diff
@@ -60,8 +60,7 @@ def link_params
     params.require(:link).permit(
       :url,
       :name,
-      :start_day, :start_month, :start_year,
-      :end_day, :end_month, :end_year,
+      :day, :month, :year,
       :year,
```
maxf: `year` and `quarter` are redundant, as they could be derived from day/month/year and the frequency of the dataset. It would make for a cleaner model to remove this redundancy. Year is trivial, you just need helper functions to convert quarters to date and vice-versa.

PR #407, opened 2017-11-06
Make sure that organisation data is indexed when syncing datasets

TobyRet:
- When syncing legacy make sure org data is indexed too!
- Refactor sync task
- Add test coverage for when there are no datasets to import
- Add more logging to aid debugging on staging and prod

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -86,7 +86,7 @@ namespace :import do
       counter += 1
       print "Completed #{counter}\r"
 
-      MetadataTools.add_dataset_metadata(obj, orgs_cache, theme_cache)
+      MetadataTools.persist(obj, orgs_cache, theme_cache)
     end
   end
 
--- a/lib/util/legacy_dataset_sync.rb
+++ b/lib/util/legacy_dataset_sync.rb
@@ -12,15 +12,16 @@ class LegacyDatasetSync
 
   def run
     @logger.info "Importing legacy datasets...\r"
-    get_packages @host do |package|
-        begin
-          MetadataTools.add_dataset_metadata(package, @orgs_cache, @theme_cache)
-          dataset = Dataset.find_by!(uuid: package["id"])
-          LegacyPublishToElasticWorker.perform_async(dataset)
-        rescue => e
-          @logger.error e.message
-        end
-        @count += 1
+    get_legacy_datasets @host do |package|
+      begin
+        @logger.info "Attempting to save legacy dataset to postgres and elasticsearch - legacy_id: #{package["id"]}"
+        MetadataTools.persist(package, @orgs_cache, @theme_cache)
+        MetadataTools.index(package)
+        @logger.info "Legacy dataset saved - legacy_id: #{package["id"]}"
+      rescue => e
+        @logger.error e.message
+      end
+      @count += 1
     end
     @logger.info "Imported #{@count} datasets...\r"
   end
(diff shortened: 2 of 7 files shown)


lib/tasks/import.rake
```diff
@@ -86,7 +86,7 @@ namespace :import do
       counter += 1
       print "Completed #{counter}\r"
 
-      MetadataTools.add_dataset_metadata(obj, orgs_cache, theme_cache)
+      MetadataTools.save_to_postgres(obj, orgs_cache, theme_cache)
```
laurentqro: how about just `persist`?

TobyRet: @laurentqro - Yep - I can change 'save_to_postgres' to 'persist' and might as well change 'save_to_elasticsearch' to 'index'

lib/util/legacy_dataset_sync.rb
```diff
@@ -55,4 +54,8 @@ def fetch_json(url)
     @logger.error "Failed to make the request to #{url}"
     return nil
   end
+
+  def there_are(datasets)
```
laurentqro: when a method returns a boolean, it's idiomatic in Ruby to suffix the method with `?`, i.e. `there_are?(datasets)`.

Also, in this case, no need to check for the presence of datasets. If there aren't any, the block never gets called.

```
irb> [].each { |d| puts "hello" }
irb> []
```

govuklaurence: Possibly even better to return the value and do a standard `any?` on the result

TobyRet: Good point re: the '?' - will change

Interesting re: the 'each' block

I think it still makes sense to check the json structure is correct no?

i.e I want to check that I have a 'results' response before proceeding ...

```
def there_are?(datasets)
    datsets['results'] && datasets['result']['results']
end
```

That way it won't blow up if I get something unexpected ...
```
datasets['foo']['bar']
```

TobyRet: Changed to the below

lib/util/legacy_dataset_sync.rb
```diff
@@ -12,15 +12,16 @@ def initialize(orgs_cache:, theme_cache:, host:, logger:)
...
-        end
-        @count += 1
+    get_legacy_datasets @host do |package|
+      begin
+        @logger.info "Attempting to save legacy dataset to postgres and elasticsearch - legacy_id: #{package["id"]}"
+        MetadataTools.persist(package, @orgs_cache, @theme_cache)
+        MetadataTools.index(package)
+        @logger.info "Legacy dataset saved - legacy_id: #{package["id"]}"
+      rescue => e
+        @logger.error e.message
```
govuklaurence: Is it intended to skip after a failure?

lib/util/legacy_dataset_sync.rb
```diff
@@ -29,21 +30,19 @@ def run
...
   # Keep yielding recent packages until the metadata_modified and metadata_created
   # is earlier than yesterday.
-  def get_packages(server)
-    modified_url = "#{server}/api/3/action/package_search?q=metadata_modified:[NOW-1DAY%20TO%20NOW]"
-    created_url = "#{server}/api/3/action/package_search?q=metadata_created:[NOW-1DAY%20TO%20NOW]"
-
-    new_packages = fetch_json(created_url)
-    modified_packages = fetch_json(modified_url)
+  def get_legacy_datasets(server)
+    modified_datasets_url = "#{server}/api/3/action/package_search?q=metadata_modified:[NOW-1DAY%20TO%20NOW]"
```
govuklaurence: `URI::join`

spec/controllers/datasets_controller_spec.rb
```diff
@@ -22,23 +22,6 @@
...
-
-    dataset.reload
-
-    expect(dataset.title).to eq("New title")
-
-    expect(WebMock)
-      .to have_requested(:post, 'https://test.data.gov.uk/api/3/action/package_patch')
-      .once
-  end
-
```
laurentqro: yeah we're about to merge a PR where this spec is completely rewritten anyway, so not worth fixing.

lib/util/legacy_dataset_sync.rb
```diff
@@ -29,21 +30,19 @@ def run
...
-      yield pkg
-    end
+    new_legacy_datasets = fetch_json(new_datasets_url)
+    modified_legacy_datasets = fetch_json(modified_datasets_url)
 
-    modified_packages['result']['results'].each do |pkg|
-      yield pkg
+    [new_legacy_datasets, modified_legacy_datasets].each do |datasets|
+      if there_are? datasets
+        datasets['result']['results'].each do |pkg|
```
govuklaurence: As @laurentqro mentioned earlier, `each` will do nothing with an empty object, if you want to make sure you do get something when navigating down the structure, use `:fetch` with a default of `{}` instead of `:[]`.

Also in general - ask for forgiveness, not permission, would be easier to let the `IndexError` or `KeyError` happen and then handle it.

PR #413, opened 2017-11-07
Remove seemingly obsolete 'stage' logic

laurentqro:
Dataset `stage` logic introduced with https://github.com/datagovuk/publish_data_beta/commit/f7b182f4 seems to be unused.

We set the dataset's stage to `completed` when the user reaches the last step of the dataset creation process, and when a dataset is published. But the only place we check the stage of a dataset is in a view where we check whether the value is `"initialised"`.


laurentqro:
thanks @govuklaurence

PR #416, opened 2017-11-08
Remove stage column from datasets table

laurentqro:
Removes obsolete column `stage` from the `datasets` table.

Follows https://github.com/datagovuk/publish_data_beta/pull/413.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -37,6 +37,10 @@ class Dataset < ApplicationRecord
   scope :owned_by, ->(creator_id) { where(creator_id: creator_id) }
   scope :published, ->{ where(status: "published") }
 
+  def self.columns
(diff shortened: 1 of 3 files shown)


app/models/dataset.rb
```diff
@@ -37,6 +37,10 @@ class Dataset < ApplicationRecord
   scope :owned_by, ->(creator_id) { where(creator_id: creator_id) }
   scope :published, ->{ where(status: "published") }
 
+  def self.columns
+    super.reject { |c| c.name == "stage" }
+  end
```
laurentqro: Patch needed to avoid `PGError: ERROR: column "stage" does not exist`

ActiveRecord caches table columns, and uses this cache to build INSERT statements. So even if the code is not touching that column, ActiveRecord will still attempt to set it to `NULL` when saving models.

PR #418, opened 2017-11-08
Refactor legacy to beta sync

TobyRet:


app/models/legacy/server.rb
```diff
@@ -14,4 +19,12 @@ def host
   def headers
     { Authorization: ENV.fetch('LEGACY_API_KEY') }
   end
+
+  def fetch_json(url)
+    response = RestClient.get url
+    return JSON.parse(response.body)
+  rescue RestClient::ExceptionWithResponse
+    Rails.logger.error "Failed to make the request to #{url}"
+    return nil
```
govuklaurence: We're losing information here about why the request failed

govuklaurence: Because this error is being rescued it's not going to be sent to sentry -- if you don't want it to blow up the stack, you'll have to catch the error in a variable `rescue BlahError => e` and send it manually `Raven.capture_exception(e)`

laurentqro: the Sentry documentation is badly worded https://docs.sentry.io/clients/ruby/#reporting-failures. They make it sound like you don't need to manually send the error to Sentry.

govuklaurence: disagree - the two cases are distinct

PR #423, opened 2017-11-14
FIX - import organisations rake task

TobyRet:
- Fix import error (occurs when trying to parse JSON Lines file)
- Fetch legacy organisations via http rather than import from local file
- Add / consolidate logging

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -17,93 +17,72 @@ namespace :import do
 
   desc "Import organisations from legacy"
   task legacy_organisations: :environment do |_, args|
-
-    count = 1
-
+    logger = Logger.new(STDOUT)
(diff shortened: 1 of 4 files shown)


lib/tasks/import.rake
```diff
@@ -49,36 +61,36 @@ namespace :import do
...
     relationships.each do |child, parent|
       o = Organisation.find_by(name: child)
       o.parent = Organisation.find_by(name: parent)
       o.save!(validate: false)
-      print "Assigned #{count+=1} organisations...\r"
+      child_organisation_count += 1
     end
 
-    puts "\nDone"
+    logger.info "Assigned #{child_organisation_count} organisations...\r"
```
govuklaurence: are you aware that this means it will no longer display progress?

PR #433, opened 2017-11-17
Use aliases to enable zero down time when re-indexing

TobyRet:
Ready for merge!

- Index names now include a timestamp - e.g. datasets-development_20171117190442
- The 'live' index will reachable via an alias - e.g datasets-development
- When reindexing occurs:
  1. A new index is created
  2. Datasets are indexed
  3. The alias is removed from the old index and added to the new index
  4. All 'old' indexes are deleted except for the most recent 3.


lib/tasks/search.rake
```diff
@@ -45,34 +45,86 @@ namespace :search do
...
+        {index: {_id: dataset.id, data: dataset.as_indexed_json}}
+      end
     end
-  end
 
-  def prepare_records(datasets)
-    datasets.map do |dataset|
-      {index: {_id: dataset.id, data: dataset.as_indexed_json}}
+
+    def index_datasets
```
govuklaurence: would be better to pass function arguments that use instance vars imo

lib/tasks/search.rake
```diff
@@ -45,34 +45,86 @@ namespace :search do
...
+          puts 'Failed to delete alias'
+          puts e.message
+        end
+      end
+      puts 'Import done'
+    end
+
+    index_datasets
+    update_alias
+    delete_old_aliases
```
TobyRet: Currently it deletes any aliases older than a week but since we do not reindex that often perhaps I should instead keep the most recent 2 / 3 and delete any others?

lib/tasks/search.rake
govuklaurence: *than use

TobyRet: 👍

lib/tasks/search.rake
TobyRet: Will look at testing also

lib/tasks/search.rake
```diff
@@ -45,34 +45,86 @@ namespace :search do
...
-    datasets.map do |dataset|
-      {index: {_id: dataset.id, data: dataset.as_indexed_json}}
+
+    def index_datasets
+      @client.indices.create index: @new_index_name, body: {mappings: @dataset_mappings}
+      nb_dataset_processed = 0
+      Dataset.published.find_in_batches(batch_size: 50) do |datasets|
+        puts " Batching #{datasets.length} datasets (done #{nb_dataset_processed} / #{@nb_datasets_to_index})"
+        bulk_index(datasets, @new_index_name)
+        nb_dataset_processed += 50
```
govuklaurence: please add in an option to alter batching size

lib/tasks/search.rake
```diff
@@ -45,34 +45,86 @@ namespace :search do
...
+
+    def delete_old_aliases
+      aliases = @client.indices.get_aliases.keys
+      aliases.each do |alias_name|
+        begin
+          if Time.parse(alias_name.gsub(/`datasets-development_`/, '')) < 1.weeks.ago
+            @client.indices.delete index: alias_name
+            puts "Deleted alias #{alias_name}"
+          end
+        rescue=>e
```
govuklaurence: Missing some spaces here around the hashrocket

lib/tasks/search.rake
```diff
@@ -45,34 +45,86 @@ namespace :search do
...
+          puts 'Failed to delete alias'
+          puts e.message
```
govuklaurence: Use logger

lib/tasks/search.rake
```diff
@@ -45,34 +45,86 @@ namespace :search do
...
+        end
+      end
+      puts 'Import done'
+    end
+
+    index_datasets
```
govuklaurence: is there any chance of errors during these processes?

govuklaurence:
If you have the time (this is non-essential) can you profile the memory on this task and just make sure it isn't leaking anywhere? I recall some issues

TobyRet:
@govuklaurence -have slacked you a question re: memory profiling

app/services/index_deleter.rb
```diff
@@ -0,0 +1,37 @@
...
+    msg = "Failed to delete alias.\n#{e.message}"
+    logger.error msg
+    Raven.capture_error msg
+  end
+
+  private
+
+  attr_reader :client, :index_alias, :logger
+
+  def prepare_indexes_for_deletion(aliases)
```
TobyRet: Let me know if this section is confusing or should be rewritten

app/services/indexer.rb
```diff
@@ -0,0 +1,92 @@
...
+            index: 'not_analyzed'
+          },
+          uuid: {
+            type: 'string',
+            index: 'not_analyzed'
+          },
+          location1: {
+            type: 'string',
+            fields: {
+              raw: {
```
hannako: why is this raw? I'm not sure we need to be able to do a keyword search on location. See https://github.com/datagovuk/find_data_beta/pull/313

TobyRet: @hannako This has been here for a while. It is allowing an exact match on location, which is required for the location filter to work. (specifically, populating locations => if I type 'E' in the location filter field then 'England', 'England, Wales' etc should appear below).  If the requirements for the locations filter change then I imagine this mapping will require updating

app/services/alias_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+      actions: [
+        {
+          remove: {
+            index: Dataset.index_name,
+            alias: index_alias
+          }
+        }
+      ]
+    }
+  rescue
```
laurentqro: Looks like we're rescuing a very specific kind of error here. How about specifying which error we're rescuing?

app/services/alias_updater.rb
```diff
@@ -0,0 +1,50 @@
...
+        {
+          add: {
+            index: new_index_name,
+            alias: index_alias
+          }
+        }
+      ]
+    }
+  rescue => e
+    msg = "Could not update alias.\n #{e.msg}"
```
laurentqro: Does the error object have a `msg` method on it? 🤔

app/services/index_deleter.rb
```diff
@@ -0,0 +1,37 @@
...
+    # Ensure that the three most recent indexes are kept
```
laurentqro: this comment is probably a sign that you should rename your method to be more explicit about the intent.

app/services/index_deleter.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(args)
+    @index_alias = args[:index_alias]
+    @client = args[:client]
+    @logger = args[:logger]
+  end
+
+  def run
+    aliases = client.indices.get_aliases.keys
+    indexes_to_be_deleted = prepare_indexes_for_deletion aliases
+    delete indexes_to_be_deleted
```
laurentqro: nitpicking here, but the use of parentheses are more idiomatic in Ruby.

`delete(indexes_to_be_deleted)`

app/services/index_deleter.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  attr_reader :client, :index_alias, :logger
+
+  def prepare_indexes_for_deletion(aliases)
+    # Ensure that the three most recent indexes are kept
+    aliases
+      .select { |alias_name| alias_name.include? index_alias }
+      .sort_by { |alias_name| Time.parse(alias_name.gsub(/"#{index_alias}_"/, '')) }
+      .reverse
+      .slice(2...-1)
```
laurentqro: so you're reversing the array and taking the last three? How about keeping the array as is, and then taking the three first ones with the super nice https://ruby-doc.org/core-2.2.0/Array.html#method-i-take ?

laurentqro: regarding sorting, this might be overkill here, but still interesting to consider: https://ruby-doc.org/core-2.4.0/Comparable.html.

laurentqro: instead of using `select`, you might also consider the very nice https://ruby-doc.org/core-2.1.0/Enumerable.html#method-i-grep.

app/services/indexer.rb
```diff
@@ -0,0 +1,92 @@
+class Indexer
```
laurentqro: this class specialises in indexing `datasets`, so might be worth renaming this class to something like `dataset_indexer`.

app/services/reindex_service.rb
```diff
@@ -0,0 +1,24 @@
...
+    @index_deleter = args[:index_deleter]
+    @logger = args[:logger]
+  end
+
+  def run
+    logger.info "Indexing #{published_datasets_count} datasets"
+    indexer.run
+    alias_updater.run
+    index_deleter.run
+    logger.info 'Import complete'
```
laurentqro: maybe something more consistent with the phrasing on line 12? e.g. `Indexing complete`. Also, it's the reindex_service, so why not say `Reindexing complete` ?

app/services/alias_updater.rb
```diff
@@ -0,0 +1,50 @@
+class AliasUpdater
```
laurentqro: or `AliasUpdateService` to be consistent with the other service names?

PR #437, opened 2017-11-20
Zero-downtime, Blue-Green, Continuous Deploys

govuklaurence:
What this does:
* Adds `scripts/deploy.sh` that requires a few ENV vars (can be seen in the script). When invoked it will blue-green deploy a single app, such as `scripts/deploy.sh publish-data-beta-staging staging-app`
* Apps are completely rebuildable from their newly split out manifests (unfortunately required for BGD)
* Services (especially user-provided secrets/config services) are not yet easily rebuildable. I'll address this soon after we switch over to alphagov.
* You can manually BGD (if cf-blue-green-deploy plugin is installed https://github.com/bluemixgaragelondon/cf-blue-green-deploy) then you can `cf bgd publish-data-beta-staging -f staging-app-manifest.yml`
* If deploy fails for whatever reason then the old app stays active and the new one is left around for investigation.
* Metrics are only tracked on the 'active' app, but inactive ones can still throw errors.


govuklaurence:
Now ready to merge™

scripts/deploy.sh
```diff
@@ -0,0 +1,44 @@
...
+if [[ -z $CF_ENV ]]
+then
+  CF_ENV='staging'
+fi
+
+
+cf login -a $CF_API -u $CF_USER -p $CF_PASS -s $CF_SPACE
+cf add-plugin-repo CF-Community https://plugins.cloudfoundry.org
+cf install-plugin blue-green-deploy -r CF-Community -f
+cf bgd $CF_APP -f $CF_ENV-manifest.yml
```
hannako: Do you need to rename the staging-app-manifest.yml -> staging-manifest.yaml for this to work?

govuklaurence: nope :) look in .travis.yml for usage. I had to split the manifests unfortunately because otherwise this command fails to run. I left the originals so that manual deploys are still easy & familiar to do.

govuklaurence:
Oh, also, pushing to the `production` branch will automatically deploy production.

govuklaurence:
Side note: pushing to production is not allowed -- PRs only. I've already created and protected the branch HARD.

scripts/deploy.sh
maxf: A usage string would be useful, with env var descriptions.

TobyRet: Yep 'plus one' from me also - ta

govuklaurence: Added

govuklaurence:
Deleted old manifests.

scripts/deploy.sh
hannako: Does my original comment apply now you've deleted the old ones?! 🙃

govuklaurence: Um... maybe? I don't think so

hannako: But there's no staging-manifest.yml anymore right it's called staging-app-manifest.yml , so how can `CF_ENV-manifest.yml` work? Or am i being extremely dim?

PR #444, opened 2017-11-24
Fetch data dumps via http, plus small refactorings

TobyRet:


lib/tasks/import.rake
```diff
@@ -21,18 +21,14 @@ namespace :import do
     organisation_count = 0
     child_organisation_count = 0
     relationships = {}
-    host = ENV.fetch('LEGACY_HOST')
+    host = 'https://data.gov.uk/'
```
TobyRet: Note that dumps on test.data.gov.uk are not currently available. Max looking into this. Once fixed then will add an env var

lib/tasks/import.rake
```diff
@@ -70,38 +66,41 @@ namespace :import do
...
 
     # Maps the organisation UUIDs to the organisation IDs
+    logger = Logger.new(STDOUT)
     orgs_cache = Organisation.all.pluck(:uuid, :id).to_h
     theme_cache = Theme.all.pluck(:title, :id).to_h
     counter = 0
+    host = 'https://data.gov.uk/'
+    path = 'data/dumps/data.gov.uk-ckan-meta-data-latest.v2.jsonl.zip'
+    url = URI::join(host, path).to_s
+    file = download_data('latest_legacy_datasets', url, logger)
```
TobyRet: Data will be downloaded automatically, rather than making you do it yourself :-)

lib/tasks/import.rake
```diff
@@ -21,18 +21,14 @@ namespace :import do
...
     logger.info 'Processing parent organisations'
-
-    file = Tempfile.new('latest_organisations')
-    file.binmode
-    file.write(RestClient.get(url).body)
-    file.close
-
-    read_json_from_zip(file) do |obj|
+    read_json_from_zip(file, logger) do |obj|
+      # TODO - move to MetadataTools.rb
```
TobyRet: There is still a fair amount of work that needs to be done to refactor this file further, but not a high priority right now

PR #450, opened 2017-11-27
Remove publish -> legacy sync code

laurentqro:
Remove all code related to syncing datasets from Publish to Legacy.


lib/util/metadata_tools.rb
```diff
@@ -15,11 +15,8 @@ def index(obj)
   def persist(obj, orgs_cache, theme_cache)
```
hannako: remove comment

laurentqro:
@hannako @TobyRet thanks for the review. `Legacy::Server` is still used and useful for importing legacy datasets into Publish Beta, so I removed only the logic in there that was related to beta => legacy sync.

laurentqro:
@TobyRet factories are ok

PR #458, opened 2017-12-01
[DO NOT MERGE] - TImeseries bug fix. ATTN: Laurent

TobyRet:


app/services/legacy/dataset_import_service.rb
```diff
@@ -0,0 +1,307 @@
...
+    return 'other' if licence != "uk-ogl"
+    licence
+  end
+
+  def build_licence_other
+    return licence if licence != "uk-ogl"
+  end
+
+  # Converts a legacy frequency into a new-style frequency
+  def build_frequency
```
laurentqro: You've changed the behaviour here: the old code makes sure if any resource has no date, then the dataset frequency is set to `never`. Your method now omits this.

TobyRet: Hi @laurentqro - this is why some datasets were not displaying correctly - specifically datasets that had time series data and additional data. Frequencies were being mutated to 'never' even if the dataset was intentionally a time series dataset. This code is not accounting for non time series datasets though - going to investigate next

TobyRet:
Going to rebase onto @laurentqro ' changes later

PR #460, opened 2017-12-04
Fix for time series bug

TobyRet:
- Add test fixtures 
- Refactor Datasets importer 
- Fix broken tests 
- Remove MetadataTools 
- Wireup Datasets importer 
- Temporarily stub beta sync service tests


app/services/legacy/beta_sync_service.rb
```diff
@@ -0,0 +1,43 @@
...
+
+  def new_datasets
+    @legacy_server.get ENDPOINTS[:new_datasets]
+  end
+
+  def import(legacy_dataset)
+    legacy_dataset_id = legacy_dataset['id']
+    @logger.info "Attempting to save legacy dataset to postgres and elasticsearch - legacy_id: #{legacy_dataset_id}"
+    Legacy::DatasetImportService.new(legacy_dataset,@orgs_cache, @theme_cache).run
+    Legacy::DatasetIndexService.new.index(legacy_dataset['id'])
```
TobyRet: @laurentqro - Creating instances of DatasetIndexService and DatasetImportService potentially makes this class difficult to test. An option would be to inject but this will break the contact with DatasetImportService#run and require further edits as a result

spec/fixtures/legacy_dataset.json
```diff
@@ -1,119 +1,140 @@
...
+      {
+        "hash": "",
+        "description": "Resource 2 description",
+        "created": "2017-12-04T09:35:25.928982",
+        "url": "https://example.com/file.html",
+        "format": "html",
+        "id": "e12a256b-893f-44f7-ba49-8dfafe41e718"
+      }
+    ],
+    "timeseries_resources": [
```
TobyRet: I've added a timeseries and additional info resource to this dataset

spec/services/legacy/integration/beta_sync_service_spec.rb
```diff
@@ -1,7 +1,7 @@
 require 'rails_helper'
-require 'util/metadata_tools'
+# require 'util/metadata_tools'
 
-describe Legacy::LegacyToBetaSyncService do
+xdescribe Legacy::BetaSyncService do
```
TobyRet: These tests are stubbed pending further discussion on Legacy::DatasetImportService (see above)

spec/services/legacy/integration/data_importer_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+describe Legacy::DatasetImportService do
```
TobyRet: This test suite may not be required - to be removed

app/services/legacy/beta_sync_service.rb
TobyRet: @laurentqro - update - have managed to get tests working with the above implementation

TobyRet: Update - have managed to create to test the BetaSyncService with the current implementation but it relies heavily on mocks - may need to rethink this

app/services/legacy/dataset_import_service.rb
```diff
@@ -35,49 +35,85 @@ def update_or_create_dataset
...
-        rescue ArgumentError
-          datafile.start_date = Date.new(1,1,1)
-        end
-        begin
-          datafile.end_date = Date.parse(dates[1])
-        rescue ArgumentError
-          datafile.end_date = Date.new(1,1,1)
-        end
+  def create_non_timeseries_datafiles(dataset)
+    if @legacy_dataset.has_key?('individual_resources')
```
laurentqro: how about casting legacy_dataset['individual_resources'] into an array to get rid of the `if` statement?

```
Array(legacy_dataset['individual_resources']).each do |resource|
   ...
end
```

app/services/legacy/dataset_import_service.rb
```diff
@@ -35,49 +35,85 @@ def update_or_create_dataset
...
+  end
+
+  def create_datafile_date_attributes(resource)
+    dates = get_start_end_date(resource["date"])
+    start_date = Date.parse(dates[0])
+    end_date = Date.parse(dates[1])
+
+    {
+      start_date: start_date,
+      end_date: end_date,
```
laurentqro: this makes me think we should have links and docs set the dates the same way, i.e. using `day` `month`, and `year`.

laurentqro: something for another PR.

app/services/legacy/dataset_import_service.rb
```diff
@@ -35,49 +35,85 @@ def update_or_create_dataset
...
+      day: end_date.day,
+      month: end_date.month,
+      year: end_date.year
+    }
+  end
+
+  def datafile_name(resource)
+    resource['description'].strip == '' ? 'No name specified' : resource['description']
```
laurentqro: I know this was already in before, but I think this is a view concern, i.e. if the description is empty, then _show_ "No name specified" to the user. But this "No name specified" should not be stored on the model.

TobyRet:
[Legacy datasets are not displayed as time series](https://trello.com/c/V5hQes72/71-legacy-datasets-are-not-displayed-as-time-series)

app/services/legacy/dataset_import_service.rb
TobyRet: Yep good idea

spec/services/legacy/integration/data_importer_spec.rb
laurentqro: we should put everything in `dataset_import_service_spec`. All this stuff is unit testing, not integration.

TobyRet: Currently 'unit/dataset_import_service_spec' contains both integration and unit tests. 

When you call the run method, I believe you are also invoking ActiveRecord and the db.

I could move the #run tests into 'integration/dataset_import_service_spec', or may worth considering mocking ActiveRecord

app/services/legacy/dataset_import_service.rb
```diff
@@ -35,49 +35,79 @@ def update_or_create_dataset
...
-    secondary_theme  = legacy_dataset["theme-secondary"]
+    old_theme = legacy_dataset["theme-primary"]
+    secondary_theme = legacy_dataset["theme-secondary"]
     dataset.theme_id = themes_cache.fetch(old_theme, nil)
     dataset.secondary_theme_id = themes_cache.fetch(secondary_theme, nil)
     dataset.status = "published"
     dataset.save!(validate: false)
   end
 
   def create_datafiles(dataset)
```
TobyRet: Not important now, but recommend for a future PR ... the database_importer_service has a lot of public instance methods which should either be private (the private methods should be tested by calling the public one, which in this case will probably be #run) or could be moved to public interfaces on other classes.

laurentqro: yeah agreed. This PR is already massive, so let's do that in another PR. 👍

app/services/legacy/dataset_import_service.rb
```diff
@@ -35,49 +35,79 @@ def update_or_create_dataset
...
+      datafile.save!(validate: false)
     end
+  end
+
+  def create_datafile_base_attributes(resource, dataset)
+    {
+      uuid: resource["id"],
+      format: resource["format"],
+      name: datafile_name(resource),
+      created_at: dataset.created_at,
```
hannako: Isn't `created_at` a default set by active record when you create a datafile. Not sure if it's possible/wise to try and overwrite it??

lib/util/metadata_tools.rb
```diff
@@ -1,252 +0,0 @@
-module MetadataTools
```
hannako: 😢 👋

PR #475, opened 2017-12-22
Fix manifest name in deploy script

laurentqro:


maxf:
Quick note: please describe why the change, as the 'what' is usually obvious in good commits.

PR #477, opened 2018-01-02
Fix for legacy sets that have invalid start date

TobyRet:
Some legacy datasets have an invalid dates (e.g. 31/06/2016).
This causes a problem when importing (Date validation error).
To get around this we need to set datafile start and end dates to nil. This requires skipping the set :set_end_date call-back in the Link model


app/models/link.rb
```diff
@@ -1,7 +1,9 @@
 class Link < Datafile
   attr_accessor :day, :month, :year
-  
-  before_save :set_end_date
+
+  # Some legacy datafiles have invalid dates (e.g. 31/06/15).
+  # When this occurs the date attributes are empty, therefore we cannot invoke this call-back
+  before_save :set_end_date, unless: ->{ year.nil? }
```
maxf: will this work with daily datasets, or datafiles without dates (which can be saved even if they don't have a year)

hannako: not sure i understand why you're only setting end date if the year isn't nil. Rather than only setting it if the start date is nil. Can you explain why this is better?

app/services/legacy/dataset_import_service.rb
```diff
@@ -84,21 +85,36 @@ def create_resource_base_attributes(resource, dataset)
...
+      end_date: end_date
+    }
+
+    day_month_year = day_month_year_from(end_date)
+    date_attributes.merge(day_month_year)
+  end
+
+  def parse_date(date)
+    Date.parse(date)
+  rescue ArgumentError
```
hannako: Check with @laurentqro, but I thought the correct syntax for exception handling in Ruby would be to wrap the code that might raise the exception in a `begin..rescue..end` block

PR #484, opened 2018-01-03
Set dates consistently across legacy and publish datafiles

hannako:
This PR makes the way we set `start_date` and `end_date` on datafiles consistent across legacy and publish.

Previously start date and end date were being set by the `dataset_import_service`. 
However there was also a `before_save` on the link model to set `end_date`. 

This meant:
 - Imported datafiles were having end date set twice.
- Imported datafiles had both start and end dates
- Published datasets would always have a start_date of nil.

Removal of either one of start or end date on the Link model adds complexity to Find.

### DIFF
--- a/app/models/link.rb
+++ b/app/models/link.rb
@@ -4,6 +4,7 @@ class Link < Datafile
   # Some legacy datafiles have invalid dates (e.g. 31/06/15).
   # When this occurs the date attributes are empty, therefore we cannot invoke this call-back
   before_save :set_end_date, unless: ->{ year.nil? }
+  before_save :set_start_date, unless: ->{ year.nil? }
 
   validate  :validate_date_input, unless: ->{ dataset.never? }
   validates :quarter, presence: true, if: ->{ dataset.quarterly? }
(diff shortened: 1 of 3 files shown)


app/services/legacy/dataset_import_service.rb
```diff
@@ -56,8 +56,10 @@ def create_datafiles(dataset)
       datafile = Link.find_or_create_by(url: legacy_datafile["url"], dataset_id: dataset.id)
       base_attributes = create_resource_base_attributes(legacy_datafile, dataset)
       date_attributes = create_datafile_date_attributes(legacy_datafile)
+      datafile.day = date_attributes[:day]
```
TobyRet: Doesn't this do the same job?
```
datafile.assign_attributes(date_attributes)
```

hannako: Assign attributes will save the `day` to the database. The idea is to leave that to the `before_save` call back so that the way dates are saved is consistent for datafiles that are imported and those made on publish

app/models/link.rb
```diff
@@ -4,6 +4,7 @@ class Link < Datafile
   # Some legacy datafiles have invalid dates (e.g. 31/06/15).
   # When this occurs the date attributes are empty, therefore we cannot invoke this call-back
   before_save :set_end_date, unless: ->{ year.nil? }
+  before_save :set_start_date, unless: ->{ year.nil? }
```
laurentqro: @hannako How about a `set_date_range` method that computes both at once? You'd avoid passing around `date_type` as arguments and would get rid of the if statements.

Maybe something like:

```
def set_date_range
   [self.start_date, self.end_date] = compute_date_range
end

def compute_date_range
   return monthly_date_range          if dataset.monthly?
end

def monthly_date_range
   [Date.new(year.to_i, month.to_i).beginning_of_month, Date.new(year.to_i, month.to_i).end_of_month]
end
```

maxf: That would indeed look slightly better. However, since the plan is to remove start_date altogether (which we haven't yet for the reasons explained above) it'll be easier, and diff will be clearer. Make sense, @laurentqro ?

laurentqro: 👍

PR #489, opened 2018-01-05
Max/documentation flag

maxf:
Currently when datafiles are sent to ES, it doesn't differentiate between data links and documents. That's because the Type field of Datafile (set to Link or Doc) isn't converted to JSON. 
Best would be to separate out Links and Docs and remove Datafile as their common ancestor. In the meantime with this we're setting the (previously unused) `documentation` flag. Find will be modified accordingly.

### DIFF
--- a/app/controllers/datasets/docs_controller.rb
+++ b/app/controllers/datasets/docs_controller.rb
@@ -1,3 +1,4 @@
+# coding: utf-8
 class Datasets::DocsController < ApplicationController
   before_action :set_dataset, only: [:index, :new, :create, :edit, :update, :confirm_delete, :destroy]
   before_action :set_doc,     only: [:edit, :update, :confirm_delete, :destroy]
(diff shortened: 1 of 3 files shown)


app/controllers/datasets/docs_controller.rb
```diff
@@ -12,6 +13,7 @@ def new
 
   def create
     @doc = @dataset.docs.build(doc_params)
+    @doc.documentation = true
```
TobyRet: I get what you are trying to do here but I wonder if having an attribute of 'type' could be worth considering. That way, it's extendable ...

e.g type='document'

Single Table inheritance would automate this also ...

maxf: Yes, this is a temporary measure. 'documentation' isn't great, but it's already in the datafile model, apparently unused. Agree with you on improving, but probably later

hannako:
@maxf Can this PR be closed now?

PR #492, opened 2018-01-16
Set up Publish and its related services to run inside Docker containers

laurentqro:
This PR sets up Publish and its related services to run inside Docker containers.

Some of the benefits:

* a new developer joining the team just has to install Docker and run `docker-compose up` and `docker-compose run web rake db:create db:migrate db:seed` to have their development environment ready

* all developers work in an identical environment

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+.git
+.dockerignore
+.byebug_history
+log/*
+tmp/*
+node_modules
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,7 +16,3 @@ RUN bundle install
 
 # Copy the contents of the Rails app folder into the container
 COPY . /usr/src/app
-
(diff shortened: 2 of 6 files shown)


laurentqro:
@hannako need to make a change to how I build the images, (i.e. revert to using something like `build: .` here: https://github.com/alphagov/datagovuk_publish/pull/492/commits/0b9ea604877fb0ee93bba1fcf23988d550de6ee5#diff-4e5e90c6228fd48698d074241c2ba760L4), and then document that the user would need to replace the `.` with the relative paths to the app folders.

cc @andrewgarner

benthorner:
As a new developer, I thought the setup was pretty simple - do we need this?

laurentqro:
Set up is indeed simple, no need for this. I did this as an experiment during firebreak.

PR #493, opened 2018-01-17
Refactor models (Datafile, Link, Doc)

TobyRet:
This PR makes changes to some models in Publish in order more accurately reflect the relationships they trying to represent.

**Before**
Link and Doc inherited from Datafile

**Now**
Doc and Datafile inherit from Link

https://trello.com/c/zPPqU6pz/114-refactoring-publish-models

**Note**
This pr makes some changes to how links (formerly datafiles) are indexed. @hannako will be making the necessary changes to Find
That you need to re-run migrations for changes to take effect


TobyRet:
Closing and reopening in order to activate Trello web hook

app/controllers/datasets/datafiles_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def set_dataset
+    @dataset = Dataset.find_by(uuid: params[:uuid])
+  end
+
+  def set_datafile
+    @datafile = Datafile.find(params[:id])
+  end
+
+  def link_params
```
laurentqro: `datafile_params`?

app/models/dataset.rb
```diff
@@ -42,8 +42,8 @@ def is_readonly?
     end
   end
 
-  def datafiles
-    links + docs
+  def links
```
laurentqro: would it possible to have `has_many :links` instead of this?

app/controllers/datasets/datafiles_controller.rb
TobyRet: Adding to next PR

app/models/dataset.rb
TobyRet: Adding to next PR

PR #497, opened 2018-01-18
Add unit tests for models

TobyRet:
Add test coverage for Link and Datafile models. No test required for Doc model as it inherits from Link and has no additional behaviour.

https://trello.com/c/NTJs5IJC/125-model-unit-tests-link-datafile-doc


spec/models/datafile_spec.rb
```diff
@@ -0,0 +1,161 @@
...
+        @datafile.end_date = nil
+
+        @datafile.save(validate: false)
+
+        expect(Datafile.last.end_date).to be_nil
+      end
+
+      it 'does not set a start date if year is nil' do
+        @datafile.year = nil
+        @datafile.start_date = nil
```
laurentqro: should this be `@datafile.start_date = Time.now` or something? Not too sure what we're testing here.

PR #502, opened 2018-01-22
Refactor dataset importer

laurentqro:
Mostly extracted local variable to a method, removed arguments where we could use the available `dataset` reader method and refactored for clarity.

### DIFF
--- a/app/services/legacy/dataset_import_service.rb
+++ b/app/services/legacy/dataset_import_service.rb
@@ -11,72 +11,70 @@ class Legacy::DatasetImportService
   def run
     update_or_create_dataset
     create_inspire_dataset(dataset.id) if dataset.dataset_type == 'inspire'
-    create_resources(dataset)
+    create_datafiles
+    create_documents
   end
 
   def update_or_create_dataset
-    dataset.legacy_name = legacy_dataset["name"]
-    dataset.title = legacy_dataset["title"]
-    dataset.summary = generate_summary(legacy_dataset["notes"])
-    dataset.description = legacy_dataset["notes"]
-    dataset.organisation_id = orgs_cache[legacy_dataset["owner_org"]]
-    dataset.frequency = build_frequency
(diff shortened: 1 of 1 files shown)


app/services/legacy/dataset_import_service.rb
```diff
@@ -11,13 +11,14 @@ def initialize(legacy_dataset, orgs_cache, themes_cache)
   def run
     update_or_create_dataset
     create_inspire_dataset(dataset.id) if dataset.dataset_type == 'inspire'
-    create_resources(dataset)
+    create_datafiles
+    create_documents
   end
 
   def update_or_create_dataset
```
TobyRet: @laurentqro (def) update_or_create_dataset is declared twice(?)

app/services/legacy/dataset_import_service.rb
```diff
@@ -36,47 +37,76 @@ def update_or_create_dataset
...
+    dataset.theme_id = build_theme_id
+    dataset.secondary_theme_id = build_secondary_theme_id
     dataset.status = "published"
     dataset.save!(validate: false)
   end
 
-  def create_resources(dataset)
-    create_datafiles(dataset)
-    create_documents(dataset)
+  def update_or_create_dataset
```
TobyRet: already declared?

TobyRet:
I'm not clear on how the changes are speeding up the import process ... I must be missing something!

app/services/legacy/dataset_import_service.rb
```diff
@@ -36,47 +37,76 @@ def update_or_create_dataset
...
+    if dataset.persisted?
+      dataset.update_columns(dataset_attributes)
```
TobyRet: ... ah ... I think this is what is speeding up the process? It's making a direct call to the db

app/services/legacy/dataset_import_service.rb
laurentqro: good catch, fixed

PR #503, opened 2018-01-23
Fix organisation importer

laurentqro:
`#update_columns` was failing silently because we were setting an invalid value for `ancestry`. Reverting to a more straighforward and simple way. It will be slower, but for 1400 something records, it's no big loss. Better to concentrate our efforts to speed up imports of datasets.

### DIFF
--- a/app/services/legacy/organisation_import_service.rb
+++ b/app/services/legacy/organisation_import_service.rb
@@ -6,11 +6,7 @@ class Legacy::OrganisationImportService
   end
 
   def run
-    if organisation.persisted?
-      organisation.update_columns(organisation_attributes)
-    else
-      organisation.update_attributes(organisation_attributes)
-    end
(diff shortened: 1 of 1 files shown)


app/services/legacy/organisation_import_service.rb
```diff
@@ -6,11 +6,7 @@ def initialize(legacy_organisation)
   end
 
   def run
-    if organisation.persisted?
-      organisation.update_columns(organisation_attributes)
-    else
-      organisation.update_attributes(organisation_attributes)
-    end
+    organisation.update_attributes(organisation_attributes)
```
TobyRet: @laurentqro - should we be using ! here and logging the exception?

laurentqro: could do, yeah

PR #505, opened 2018-01-24
Dashboard of what data is present in the DB

laurentqro:
screen shot 2018-02-07 at 2 50 55 pm

https://trello.com/c/9T5RNLeN/142-admin-reports-on-find


app/controllers/home_controller.rb
```diff
@@ -4,17 +4,24 @@ def index
     home_path_for_user
   end
 
-  # FIX: Temporary controller, remove me when no longer required
```
laurentqro: Temporary ™️

TobyRet: ha!

app/controllers/home_controller.rb
```diff
@@ -4,17 +4,26 @@ def index
...
   def quality
     @scores = QualityScore.all.order(median: :desc, highest: :desc)
   end
 
+  def dashboard
```
TobyRet: Any particular reason why the dashboard is being managed by the HomeController?

app/controllers/home_controller.rb
```diff
@@ -4,17 +4,26 @@ def index
...
+    @datasets_count = Dataset.count
+    @datafiles_count = Link.count
+    @publishers_count = Organisation.count
+    @published_datasets_count = Dataset.published.count
+    @draft_datasets_count = Dataset.draft.count
+    @unpublished_datasets_count = Dataset.left_outer_joins(:datafiles).where(links: { id: nil } ).count # i.e. datasets with no datafiles
```
TobyRet: Perhaps move to a class method called 'datasets_with_no_datafiles' ?

laurentqro: good idea

app/controllers/home_controller.rb
laurentqro: I agree, it's weird. This is ripe for a refactor.

I'd rather have a `Pages` controller for all our static pages, like we have in FIND, but this is what we have at the moment.

I thought about refactoring this too, but found it is out of the scope of this PR.

PR #506, opened 2018-01-26
Datafiles with no resource_type are datafiles too

laurentqro:
In the dataset importer, we select datafiles based on their `resource_type`. If it's `file`, they're a datafile, if it's `documentation`, they're supporting documentation.

However, it turns out not all datafiles are created equal in legacy. Some don't have a `resource_type` field. As a result, they get ignored.

This makes sure we import all datafiles. We recognize datafiles to be files if they are not documentation. This is not ideal, because it could be that some supporting documents have no `resource_type` either, but at least no one gets left behind, and we can iterate once we know more about this.

https://trello.com/c/xK99B0fA/95-missing-datafiles-on-find

### DIFF
--- a/app/services/legacy/dataset_import_service.rb
+++ b/app/services/legacy/dataset_import_service.rb
@@ -179,8 +179,8 @@ class Legacy::DatasetImportService
     get_extra("harvest_object_id").present?
   end
 
-  # Given a lax legacy date string, try and build a proper
-  # date string that we can import
+  private
+
   def get_end_date(date_string)
     # eg "1983"
     if date_string.length == 4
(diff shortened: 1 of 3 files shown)


app/services/legacy/dataset_import_service.rb
```diff
@@ -209,10 +207,8 @@ def calculate_dates_for_year(year)
     "31/12/#{year}"
   end
 
-  private
-
   def legacy_datafiles
-    resources.select{ |resource| resource['resource_type'] == 'file'}
+    resources.select{ |resource| resource['resource_type'] != 'documentation'}
```
andrewgarner: Instead of a double negative you could change this to use reject:

```ruby
resources.reject { |resource| resource['resource_type'] == 'documentation' }
```

PR #511, opened 2018-02-05
Topics

TobyRet:
- Rename Themes table to Topics (update all references)
- Index topics in Elasticsearch
- Update Elasticsearch mapping to allow filtering by topic

https://trello.com/c/Ot65Qefh/80-implement-topics-themes-l


db/migrate/20180205105808_rename_theme_id_to_topic_id.rb
```diff
@@ -0,0 +1,6 @@
+class RenameThemeIdToTopicId < ActiveRecord::Migration[5.1]
+  def change
+    rename_column :datasets, :theme_id, :topic_id
```
laurentqro: In a production environment - where we want zero-downtime - you would have to create a new column, migrate the data over, and delete the old column.

PR #512, opened 2018-02-05
Delegate link checking to background job

laurentqro:
This should make the link checking tasks faster to run and easier to debug.


laurentqro:
Have another PR for this that is more in line with the refactoring done in https://github.com/alphagov/datagovuk_publish/pull/515. Closing PR.

PR #517, opened 2018-02-07
Smarter link checker

laurentqro:
* skip link validation on save
* defer to `Ruby Net::HTTP` for checking if response is a success
* save the response error message onto the `Task` description for better reporting
* enforce a timeout for requests
* add a rake task to check all links in the DB

### DIFF
--- a/app/services/link_checker_service.rb
+++ b/app/services/link_checker_service.rb
@@ -8,10 +8,10 @@ class LinkCheckerService
   def run
     begin
       check_link
-    rescue RestClient::ExceptionWithResponse
+    rescue RestClient::ExceptionWithResponse => error
       link.broken = true
       link.save(validate: false)
-      create_broken_link_task
+      create_broken_link_task(error)
     end
   end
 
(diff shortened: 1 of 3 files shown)


app/services/link_checker_service.rb
```diff
@@ -8,24 +8,26 @@ def initialize(link)
   def run
     begin
       check_link
-    rescue RestClient::ExceptionWithResponse
-      link.update(broken: true)
-      create_broken_link_task
+    rescue RestClient::ExceptionWithResponse => error
+      link.broken = true
+      link.save(validate: false)
```
andrewgarner: Why do we need to circumvent the validation? Could we be smarter with our validation rules instead?

laurentqro: @andrewgarner as mentioned face to face, I think this is part of a wider issue, so I suggest we deal with this in a separate PR.

andrewgarner: 👍

PR #520, opened 2018-02-09
Count datasets with at least one datafile instead of counting datasets marked as published

laurentqro:
screen shot 2018-02-09 at 2 43 33 pm

### DIFF
--- a/app/views/home/dashboard.html.erb
+++ b/app/views/home/dashboard.html.erb
@@ -10,21 +10,21 @@
   <div class="column-one-third">
     <div class="data">
       <span class="data-item bold-xlarge"><%= number_with_delimiter(@datasets_count) %></span>
-      <span class="data-item bold-xsmall">total datasets</span>
+      <span class="data-item bold-xsmall"><%= 'dataset'.pluralize(@datasets_count) %> in total</span>
     </div>
   </div>
 
   <div class="column-one-third">
     <div class="data">
       <span class="data-item bold-xlarge"><%= number_with_delimiter(@datasets_published_with_datafiles_count) %></span>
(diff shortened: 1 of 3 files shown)


maxf:
@laurentqro  What's the 'published' flag used for?

laurentqro:
It's used as the opposite of `draft`

app/views/home/dashboard.html.erb
```diff
@@ -10,21 +10,21 @@
...
-      <span class="data-item bold-xsmall">published datasets</span>
+      <span class="data-item bold-xsmall">datasets published with datafiles</span>
     </div>
   </div>
 
   <div class="column-one-third">
     <div class="data">
       <span class="data-item bold-xlarge"><%= number_with_delimiter(@unpublished_datasets_count) %></span>
-      <span class="data-item bold-xsmall">datasets with no datafiles</span>
+      <span class="data-item bold-xsmall">datasets published with no datafiles</span>
```
andrewgarner: This is confusing. The variable says `@unpublished_datasets_` but the suffix says "datasets published".

laurentqro: I agree, "unpublished" is an awful name for a dataset published with no datafile, but apparently it's what everyone calls it in the legacy system. ** sigh **

I am happy to just say "published with no datafiles" everywhere. I'll make the change.

andrewgarner: The variable name implies it is a count of datasets that have not been published. However, the suffix says "datasets published".

andrewgarner: Cool! I guess we should try and introduce a domain language that's easier to understand in this brave new world :)

PR #521, opened 2018-02-13
Fix Cloud Foundry manifests (running Sidekiq)

TobyRet:
When running Sidekiq you need to pass in it's configuration file in order to correctly process jobs

Trello: https://trello.com/c/Pl95C7Kt/208-sidekiq-bug-data-import-not-working


andrewgarner:
See https://github.com/mperham/sidekiq/blob/v5.0.4/lib/sidekiq/cli.rb#L363-L365

PR #527, opened 2018-02-22
Remove Themes

TobyRet:
We are using 'topics' instead in production


db/migrate/20180222152356_remove_theme_id_from_dataset.rb
```diff
@@ -0,0 +1,6 @@
+class RemoveThemeIdFromDataset < ActiveRecord::Migration[5.1]
+  def change
+    remove_column :datasets, :theme_id
+    remove_column :datasets, :secondary_theme_id
```
laurentqro: for zero downtime:

```
class Dataset
  def self.columns
    super.reject { |c| c.name == "theme_id" || c.name == "secondary_theme_id" }
  end
end
```

PR #532, opened 2018-02-27
Make short_id's non random

hannako:
Trello card: https://trello.com/c/YJEH4PC2/258-make-uuids-not-random

When we import datasets from legacy dumps, we want the short_id's that are used in the URL's to always be the same. 

Currently:
- A single dataset has a different url in each environment. 
- If we have to reimport from a dump whilst the service is live, all the URL's will break. 

This PR implements the creation of short_id's that are based on the legacy UUID, rather than generating a random UUID each time.

### DIFF
--- a/spec/models/dataset_spec.rb
+++ b/spec/models/dataset_spec.rb
@@ -111,10 +111,32 @@ describe Dataset do
     expect(dataset.last_updated_at).to eq last_updated_at
   end
 
-  describe 'after initializing' do
-    it "generates a unique short_id upon initialising" do
-      dataset = Dataset.new
-      expect(dataset.short_id).to_not be_nil 
+  context 'when creating an imported legacy dataset' do
+    it "generates a unique deterministic short_id based on the legacy UUID" do
+      legacy_id = 'abcdef123456'
+      short_id = Digest::SHA256.hexdigest(legacy_id)[0..6]
+
+      imported_dataset = FactoryGirl.create(:dataset, uuid: legacy_id)
+      expect(imported_dataset.short_id).to eq short_id
+      imported_dataset.destroy
+
+      imported_dataset = FactoryGirl.create(:dataset, uuid: legacy_id)
+      expect(imported_dataset.short_id).to eq short_id
+    end
+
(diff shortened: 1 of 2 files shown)


spec/models/dataset_spec.rb
```diff
@@ -111,10 +111,32 @@
...
+      imported_dataset.destroy
+
+      imported_dataset = FactoryGirl.create(:dataset, uuid: legacy_id)
+      expect(imported_dataset.short_id).to eq short_id
+    end
+
+    it "continues to generate short_ids until it has found a unique one" do
+      legacy_id = '123456abcdef'
+      3.times do
+        FactoryGirl.create(:dataset, uuid: legacy_id)
```
andrewgarner: How can we have multiple legacy datasets that have the same ID? Should they not be unique?

hannako: We can't. Was just a quick way to test that the generate_short_id function would loop if there was a clash.

andrewgarner: Okay. In that case I'm not sure we need the seed. We'd want an exception to be raised if the short_id already exists since there should never be more than one dataset with the same uuid.

hannako: We skip all the validations on the import so we don't bomb out mid way through. 

I think that's why previously there was the while loop to check if the candidate_short_id is unique, and if not then a new one is generated.

andrewgarner:
I ran the following script to see if there were any duplicates in the ckan exports:

```ruby
require 'json'

puts File
       .foreach('data.gov.uk-ckan-meta-data-latest.v2.jsonl')
       .map { |line| JSON.parse(line)['id'] }
       .group_by { |id| id }
       .select { |k, v| v.size > 1 }
       .size

# => 0
```

hannako:
@andrewgarner not sure I follow. There can’t be duplicate id’s from ckan can there? The check for duplicates in the short_id was a request from Jonathan when Toby originally implemented it, because we were shortening the URL down to a 6 character short id, when it was previously the full uuid and he wanted to be certain we’d never get a clash. Or am I totally missing your point?

PR #536, opened 2018-03-09
Refactor logging

andrewgarner:
- Switches the application over to use common config for Sidekiq in GOV.UK applications – this gets us logs formatted correctly for our logging tools.

- Ensures logging information is correctly propagated when we sync data with legacy.

See https://trello.com/c/4rGs48M2


andrewgarner:
@hannako the rake task now uses `Rails.logger` so you'll need to tail `log/development.rb` for log output from `BetaUpdateWorker`.

PR #544, opened 2018-04-03
Updates the readme with more info for set up.

rossjones:
Specifies where the developer can get some sample data from. This isn't
the optimal way to do this (rather we should take dumps from staging)
but it should allow people to get up and running if they don't have
access to our env.


benthorner:
Is this PR superseded with the one we're looking at now?

benthorner:
This is superseded by https://github.com/alphagov/datagovuk_publish/pull/570.

PR #549, opened 2018-04-10
Install brakeman so we can look for common security issues

rossjones:


benthorner:
Could we make it part of the default 'rake' command?

rossjones:
I think so, we'll also probably be sensible to run on travis as well in a later PR.
Will look at adding to take today.

rossjones:
We also need to go through and fixing everything it's identified as well.  Was hoping this could be just adding it, and another PR with the fixes to existing issues could include adding it to rake.

benthorner:
Is there a benefit to adding it without running it?

rossjones:
There's a benefit to a sequence of smaller PRs :)

benthorner:
Hmm...I'm not so sure - more PRs = more overhead, no? I'd really like to see us using something at the time we add it. Otherwise it could get forgotten about, like another refactoring.

rossjones:
I find they're easier to merge when they're small.

I'll take a look at fixing all the issues it raises and adding it to the default rake after the weekend.

benthorner:
Thinking about the way we did the govuk-lint PRs, I think there's a precedent for fixing the issues raised by Brakeman in this PR. Looking at it another way, in the past I've seen the existence of a .ignore file turning into its own precedent for ignoring new issues people didn't feel like dealing with. I'm up for taking a look at some of them, as it would be useful knowledge to avoid making these kinds of mistakes when Brakeman isn't there to check for them.

rossjones:
@benthorner Feel free to take a look, I think though they are false positives, which I believe is the reason brakeman supports ignore files.

benthorner:
@rossjones I've finished if you want to take another look :-)

app/models/dataset.rb
```diff
@@ -26,7 +24,7 @@ class Dataset < ApplicationRecord
 
   validates :frequency, inclusion: { in: %w(daily monthly quarterly annually financial-year never irregular) },
     allow_nil: true # To allow creation before setting this value
-  validates :title, presence: true, format: { with: TITLE_FORMAT }
+  validates :title, presence: true
```
rossjones: Are we not losing functionality here?

benthorner: I checked this with Jon - as long as it's present he doesn't mind. It was actually checking the string had three characters 'somewhere', which is not what we want and partly why Brakeman flagged it in the first place :-).

nicholsj: We can decide if we need more specific validation on title etc later (eg character count). It sounds like it wasn't doing much useful anyway?

spec/models/dataset_spec.rb
```diff
@@ -12,17 +12,6 @@
     expect(dataset.errors[:summary]).to include "Please provide a summary"
   end
 
-  it "validates title format" do
-    dataset = Dataset.new(title: "[][]")
```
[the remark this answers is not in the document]
benthorner: OK. I didn't connect titles and slugs when I made the change. @nicholsj is a valid title just a string of words (i.e. no special characters except whitespace)?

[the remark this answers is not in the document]
nicholsj: @rossjones @benthorner Ok, I'm going to have to leave it to devs to agree a way forward on this. As it was explained to me, it was simply checking that the title had 3 characters - and that seemed pointless to me. 

Is this another case where we should generate a new card to explore/fix an issue, rather than wrapping it up in another PR which means it gets blocked?

app/models/dataset.rb
benthorner: Not really - it was hiding the fact we weren't doing it properly. On the other hand, it would now save a dataset with an invalid slug if you entered '[]' as the title. I guess that's unlikely.

PR #555, opened 2018-04-13
Add the INSPIRE access constraints field to DB

rossjones:
The access constraints from the INSPIRE metadata was missing, and so
this PR adds a migration to put it back as a text block.

Most constraints fall into one of three categories:

* Plain text
* A URL
* Paragraphs and paragraphs of text

Changes to the import/model will come in a separate PR with a rake task
for migrating just that data that is missing.

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -10,7 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2017071231151258) do
+ActiveRecord::Schema.define(version: 2018041310444200) do
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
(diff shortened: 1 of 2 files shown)


db/schema.rb
```diff
@@ -221,6 +223,13 @@
     t.index ["organisation_id"], name: "index_quality_scores_on_organisation_id"
   end
 
+  create_table "spatial_ref_sys", primary_key: "srid", id: :integer, default: nil, force: :cascade do |t|
```
andrewgarner: I think this is a product of the `postgis` extension.

rossjones: Think it included the extensions that are in the production database from my last run of replication.sh - will squash and re-push

andrewgarner: 👍

PR #556, opened 2018-04-13
Adds a rake task to import a single dataset

rossjones:
Adds a new rake task `import:single_legacy_dataset` which, when provided
with a legacy short-name will:

* Retrieve the metadata from the legacy API
* Update the dataset record in publish
* Update the search index

During the DB update, it now sets the access_constraints field on the
InspireDataset record.

https://trello.com/c/T3K2C1lD/349-access-constraints-missing-from-inspire-metadata

### DIFF
--- a/app/services/legacy/dataset_index_service.rb
+++ b/app/services/legacy/dataset_index_service.rb
@@ -18,7 +18,8 @@ class Legacy::DatasetIndexService
                                                 type: ::Dataset.__elasticsearch__.document_type,
                                                 id: dataset.id
                                               })
-    rescue Elasticsearch::Transport::Transport::Errors::NotFound
+    rescue Elasticsearch::Transport::Transport::Errors::NotFound => e
+      logger.warn e.message
     end
   end
 end
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -3,6 +3,8 @@ require 'csv'
 require 'zip'
 require 'rest-client'
 
+LEGACY_SHOW_API = 'https://data.gov.uk/api/3/action/package_show'
+
 namespace :import do
 
   desc "Import locations from a CSV file"
@@ -54,7 +56,7 @@ namespace :import do
 
     begin
       api_parameters = { params: { id: args.legacy_shortname } }
-      api_response = RestClient.get 'https://data.gov.uk/api/3/action/package_show', api_parameters
+      api_response = RestClient.get LEGACY_SHOW_API, api_parameters
     rescue RestClient::ExceptionWithResponse => e
       logger.error "Request to API to retrieve #{args.legacy_shortname} responded with: #{e.response.code}"
       next


lib/tasks/import.rake
```diff
@@ -44,10 +43,38 @@ namespace :import do
...
+
+  desc "Import a single legacy dataset from the legacy API"
+  task :single_legacy_dataset, [:legacy_shortname] => :environment do |_, args|
+    logger = Rails.logger
+
+    begin
+      api_parameters = { params: { id: args.legacy_shortname } }
+      api_response = RestClient.get 'https://data.gov.uk/api/3/action/package_show', api_parameters
+    rescue RestClient::ExceptionWithResponse => e
+      logger.info "Request to API to retrieve #{args.legacy_shortname} responded with: #{e.response.code}"
```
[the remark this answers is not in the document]
rossjones: Possibly. I'll go with that when I work out why tests failing.

lib/tasks/import.rake
```diff
@@ -44,10 +43,38 @@ namespace :import do
...
+    rescue RestClient::ExceptionWithResponse => e
+      logger.error "Request to API to retrieve #{args.legacy_shortname} responded with: #{e.response.code}"
+      next
+    end
+
+    legacy_dataset = JSON.parse(api_response).fetch('result')
+    Legacy::DatasetImportService.new(legacy_dataset, organisation_cache, topic_cache).run
+
+    indexer = Legacy::DatasetIndexService.new
+    indexer.remove_from_index(legacy_dataset['id'])
```
benthorner: Would it be better to create before we remove in case the latter goes wrong?

lib/tasks/import.rake
```diff
@@ -44,10 +43,38 @@ namespace :import do
...
     logger.info 'Import complete'
   end
+
+  desc "Import a single legacy dataset from the legacy API"
+  task :single_legacy_dataset, [:legacy_shortname] => :environment do |_, args|
+    logger = Rails.logger
+
+    begin
+      api_parameters = { params: { id: args.legacy_shortname } }
+      api_response = RestClient.get 'https://data.gov.uk/api/3/action/package_show', api_parameters
```
benthorner: I usually try to put long strings as constants. Could we do that here?

lib/tasks/import.rake
rossjones: I'm not sure, the rake task is only for a specific use (backfillint missing data) so as long as it is in the database, we can reindex anytime.  Plus, don't want to remove the one we just added by mistake.  Without removing first, we end up with two entries.

app/services/legacy/dataset_index_service.rb
```diff
@@ -8,4 +8,17 @@ def index(legacy_dataset_id)
...
+  def remove_from_index(legacy_dataset_id)
+    dataset = Dataset.find_by!(uuid: legacy_dataset_id)
+
+    begin
+      dataset.__elasticsearch__.delete_document({
+                                                index: ::Dataset.__elasticsearch__.index_name,
+                                                type: ::Dataset.__elasticsearch__.document_type,
+                                                id: dataset.id
+                                              })
+    rescue Elasticsearch::Transport::Transport::Errors::NotFound
```
benthorner: Does this mean we don't care about failed removals? Perhaps we could add a log just to say there was a failure.

lib/tasks/import.rake
benthorner: OK. I was missing the context that we had a 'live backup' if it goes wrong. Duplication in the new stack is a more important issue - all good :-)

app/services/legacy/dataset_index_service.rb
rossjones: We generally don't (because it might not be there and we don't want to round-trip), but I'll add a .warn

PR #558, opened 2018-04-17
Fix licence handling 

rossjones:
This PR changes the handling of licences during import so that we have 4 new fields that we can use in bringing Find up to date.

The new fields (licence_code, licence_title, licence_url and licence_custom) will be set on import (either a full import or as a result of `rake import:single_legacy_dataset[environmental-pollution-incidents]`).  

The existing licence fields (licence and licence_other) can continue to be used until the migration is complete.

NB. This does not change the user-interface of Publish, this should be another PR.

https://trello.com/c/EwW9h6EU/12-some-licenses-missing-from-find-dataset-display-l

### DIFF
--- a/app/models/licence.rb
+++ b/app/models/licence.rb
@@ -1,0 +1,25 @@
+require 'ostruct'
+
+class Licence
+  LicenceInfo = Struct.new(:title, :url)
+
+  LICENCES = {
+    "uk-ogl" => LicenceInfo.new("Open Government Licence", "http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/"),
+    "cc-nc" => LicenceInfo.new("Creative Commons Non-Commercial (Any)", "http://creativecommons.org/licenses/by-nc/2.0/"),
+    "cc-zero" => LicenceInfo.new("Creative Commons CCZero", "http://www.opendefinition.org/licenses/cc-zero"),
+    "cc-by-sa" => LicenceInfo.new("Creative Commons Attribution Share-Alike", "http://www.opendefinition.org/licenses/cc-by-sa"),
+    "cc-by" => LicenceInfo.new("Creative Commons Attribution", "http://www.opendefinition.org/licenses/cc-by"),
+    "odc-odbl" => LicenceInfo.new("Open Data Commons Open Database License (ODbL)", "http://www.opendefinition.org/licenses/odc-odbl"),
+    "notspecified" => LicenceInfo.new("License Not Specified", nil),
+    "other-pd" => LicenceInfo.new("Other (Public Domain)", nil),
+    "other-open" => LicenceInfo.new("Other (Open)", nil),
+    "other-closed" => LicenceInfo.new("Other (Not Open)", nil),
+    "other-nc" => LicenceInfo.new("Other (Non-Commercial)", nil),
+    "odc-pddl" => LicenceInfo.new("Open Data Commons Public Domain Dedication and License (PDDL)", "http://www.opendefinition.org/licenses/odc-pddl"),
+    "odc-by" => LicenceInfo.new("Open Data Commons Attribution License", "http://www.opendefinition.org/licenses/odc-by"),
+  }
+
+  def self.lookup(licence_id)
+    LICENCES.fetch(licence_id, LicenceInfo.new)
+  end
+end
--- a/app/models/licence.rb
+++ b/app/models/licence.rb
@@ -1,25 +1,23 @@
-require 'ostruct'
-
 class Licence
   LicenceInfo = Struct.new(:title, :url)
 
   LICENCES = {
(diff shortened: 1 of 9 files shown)


andrewgarner:
We'll need to change the frontend as this is a breaking change.

rossjones:
Yes it will.

We can add the hash that maps 
```
{
  "uk-ogl": { title: "Open Gov Licence", url: "http:....."},
  "cc-by": { title: "CC-BY", url: "http:....."}
}
```
so we don't have to index that data to find, if we do that first, can we avoid a search-index-apocalypse?

app/models/licence.rb
```diff
@@ -0,0 +1,25 @@
+require 'ostruct'
```
andrewgarner: You shouldn't need this if you're using Struct.

app/models/licence.rb
```diff
@@ -0,0 +1,25 @@
...
+  LicenceInfo = Struct.new(:title, :url)
+
+  LICENCES = {
+    "uk-ogl" => LicenceInfo.new("Open Government Licence", "http://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/"),
+    "cc-nc" => LicenceInfo.new("Creative Commons Non-Commercial (Any)", "http://creativecommons.org/licenses/by-nc/2.0/"),
+    "cc-zero" => LicenceInfo.new("Creative Commons CCZero", "http://www.opendefinition.org/licenses/cc-zero"),
+    "cc-by-sa" => LicenceInfo.new("Creative Commons Attribution Share-Alike", "http://www.opendefinition.org/licenses/cc-by-sa"),
+    "cc-by" => LicenceInfo.new("Creative Commons Attribution", "http://www.opendefinition.org/licenses/cc-by"),
+    "odc-odbl" => LicenceInfo.new("Open Data Commons Open Database License (ODbL)", "http://www.opendefinition.org/licenses/odc-odbl"),
+    "notspecified" => LicenceInfo.new("License Not Specified", nil),
```
andrewgarner: I think you can just ommit the last argument and it will default to `nil`.

app/models/licence.rb
```diff
@@ -0,0 +1,25 @@
+require 'ostruct'
+
+class Licence
+  LicenceInfo = Struct.new(:title, :url)
+
+  LICENCES = {
```
andrewgarner: How about we sort this list alphabetically to make the code a bit easier to scan?

app/models/licence.rb
```diff
@@ -0,0 +1,25 @@
...
+    "notspecified" => LicenceInfo.new("License Not Specified", nil),
+    "other-pd" => LicenceInfo.new("Other (Public Domain)", nil),
+    "other-open" => LicenceInfo.new("Other (Open)", nil),
+    "other-closed" => LicenceInfo.new("Other (Not Open)", nil),
+    "other-nc" => LicenceInfo.new("Other (Non-Commercial)", nil),
+    "odc-pddl" => LicenceInfo.new("Open Data Commons Public Domain Dedication and License (PDDL)", "http://www.opendefinition.org/licenses/odc-pddl"),
+    "odc-by" => LicenceInfo.new("Open Data Commons Attribution License", "http://www.opendefinition.org/licenses/odc-by"),
+  }
+
+  def self.lookup(licence_id)
```
andrewgarner: We should use `licence_code` to be consistent with the new attribute on `Dataset`.

app/services/legacy/dataset_import_service.rb
```diff
@@ -158,11 +164,20 @@ def build_summary(notes)
...
   end
 
   def build_licence_other
-    get_extra('licence')
+    return nil if licence.blank?
+    return licence if licence != "uk-ogl"
+  end
+
+  def build_licence_code
+    code = legacy_dataset["license_id"]
```
andrewgarner: Rails has http://api.rubyonrails.org/classes/Object.html#method-i-presence which you could maybe use here.

```rb
def build_licence_code
  legacy_dataset["license_id"].presence
end
```

andrewgarner: You should also use the newly introduced `licence` method. Maybe you could use `presence` in that method?

```rb
def build_licence_code
  licence
end

def licence
  legacy_dataset["license_id"].presence
end
```

app/services/legacy/dataset_import_service.rb
```diff
@@ -43,6 +45,10 @@ def dataset_attributes
       location1: build_location,
       licence: build_licence,
       licence_other: build_licence_other,
+      licence_code: build_licence_code,
+      licence_title: licence_info.title,
+      licence_url: licence_info.url,
+      licence_custom: get_extra("licence", default: nil),
```
andrewgarner: Did you look at the side effect of setting the default to `nil` instead of just here? It seems strange to to have some attributes blank and then this one `nil`.

rossjones: I'm not overly keep on using "" to represent not present, but I guess if it's what we're doing everywhere else.

rossjones:
Updated in response to review.

PR #559, opened 2018-04-18
Increase log level to avoid redundant spam

benthorner:
My experience of rails apps is that excessive logging also degrades
performance of the application server.

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -49,7 +49,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
 
   # Prepend all log lines with the following tags.
   config.log_tags = [ :request_id ]
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -49,7 +49,7 @@ Rails.application.configure do
 
   # Use the lowest log level to ensure availability of diagnostic information
   # when problems arise.
-  config.log_level = :debug
+  config.log_level = :info
 
   # Prepend all log lines with the following tags.
   config.log_tags = [ :request_id ]


benthorner:
In my previous company we would set it to warn unless there was a specific need for more logging, which there isn't here. The problem with :info is that it will log every request - can we set a better trend here?

rossjones:
I'm a bit worried about losing information that we need need to track down issues in future.  We could be more careful where we use logger.info other than what Rails already logs.

benthorner:
So if the request fails rails will still log it. But we really don't need to be logging all the successful requests - filtering them out will actually make the problematic ones easier to read ;-)

andrewgarner:
Changing the log level to make information easier to find is the wrong approach. Logit is our chosen solution to address that need. We should only change the log level if the information we log is overly verbose.

We also need to be able to see when things are working, not just when they fail. A log level of `info` gives us that.

benthorner:
Could you share an example of when we need to see successful requests in the logs? I recall in my previous company we thought we needed them and then didn't miss them at all when we turned them off. It seems a shame to use more resources unless we have a specific use case.

rossjones:
We want to know who is hammering every /dataset/* URL and at what sort of rate.

Logit allows us to filter (and facets with that filter applied),  and we've got a short retention period.   This was very useful when Performance Platform went down and we didn't know why, we were able to identify specific user-agents that were causing the problem.  If we hadn't had logit, we would have spent twice as long trying to identify that using GA data or something similar.

andrewgarner:
- For analysing if requests are being served from a CDN or the origin servers
- For analysing request timings and identifying slow requests
- For analysing load on the application servers
- For analysing traffic patterns
- For analysing characteristics of clients

benthorner:
Fair enough - good to know :-)

PR #563, opened 2018-04-18
Add topics to publish

hannako:
[trello card #11](https://trello.com/c/scRqGVZd/11-add-topics-to-publish-workflow)

- Creates basic flow and specs for adding and editing topics on each dataset.

- Question remains on whether we should allow a dataset to be published without a topic.

Our current dataset model has `optional:true` for topic. 
And on production `Dataset.where(topic_id: nil).count => 312`
 
I discussed with JN that we would have a skip link on the topic page, as we do on the other pages of the create a dataset flow, but when the user clicks publish on the final page, they would see an error and they would then have to go back and pick a topic. This functionality is most easily implemented if we update the model to enforce having a topic.

Andrew has a better plan. I've run out of time!

### DIFF
--- a/spec/support/dataset_helper.rb
+++ b/spec/support/dataset_helper.rb
@@ -3,19 +3,19 @@ def last_updated_dataset
 end
 
 def click_change(property)
-  properties = {
-    title: 0,
-    summary: 1,
-    additional_info: 2,
-    topic: 3,
-    licence: 4,
-    location: 5,
-    frequency: 6,
-    datalinks: 7,
-    documentation: 8
-  }
-  index = properties[property]
-  all(:link, "Change")[index].click
+  properties = %i(
+    title
+    summary
+    additional_info
+    topic
+    licence
+    location
+    frequency
+    datalinks
+    documentation
+  )
+
+  all(:link, "Change")[properties.index(property)].click
 end
 
 def click_dataset(dataset)


andrewgarner:
I've introduced a validation context to @hannako's PR so that we can apply topic validations only in the context of the create a dataset form. Behaviour when a dataset remains unaffected.

app/views/datasets/show.html.erb
```diff
@@ -60,6 +60,22 @@
           <%= link_to 'Change', edit_dataset_path(@dataset.uuid, @dataset.name) %>
         </td>
     </tr>
+    <tr>
+      <tr>
+        <th class="dgu-checklist__heading">Topic</th>
+        <td>
+          <%= @dataset.topic.title if @dataset.topic %>
```
benthorner: We seem to have a mixture of `var&.field` and `var.field if var` in the code. Which one should we go with? I'm not fussed really, but we ought to be consistent.

andrewgarner: I'm happy to go with `&.` for new additions like this.

app/views/datasets/topics/_errors.html.erb
```diff
@@ -0,0 +1,17 @@
+<% if @dataset.errors.any? %>
```
benthorner: This file doesn't look specific to topics - should it go one level up so it's more likely to be re-used?

spec/features/create_dataset_spec.rb
```diff
@@ -29,13 +30,19 @@
...
+      expect(Dataset.last.topic.title).to eq(topic.title)
+
+      # PAGE 3: Licence
       choose option: "uk-ogl"
       click_button "Save and continue"
 
       expect(Dataset.last.licence).to eq("uk-ogl")
 
-      # Page 3: Location
+      # Page 4: Location
```
benthorner: Would removing the pointless page numbers be a better change here ;-)

app/views/datasets/topics/_errors.html.erb
andrewgarner: Yes it should. I created a partial here to remove duplication between the new and edit views. However there seems to be a lot of duplication between the various forms for topics, license etc that would be a good candidate to refactor out at a later date.

spec/features/edit_dataset_spec.rb
```diff
@@ -3,6 +3,7 @@
 describe 'editing datasets' do
   let(:land) { FactoryGirl.create(:organisation) }
   let!(:user) { FactoryGirl.create(:user, primary_organisation: land) }
+  let!(:business) { FactoryGirl.create(:topic) }
```
benthorner: Could we call this `:topic` instead of `business` - just so it's clear.

spec/features/edit_dataset_spec.rb
```diff
@@ -65,6 +67,16 @@
       expect(last_updated_dataset.description).to eq('a new description')
     end
 
+    it "should be able to update topic" do
+      environment = FactoryGirl.create(:topic, title: 'Environment', name: 'environment')
```
benthorner: Same here.

spec/features/create_dataset_spec.rb
andrewgarner: I'm honouring the design decisions of previous authors.

spec/support/dataset_helper.rb
```diff
@@ -7,11 +7,12 @@ def click_change(property)
     title: 0,
     summary: 1,
     additional_info: 2,
-    licence: 3,
-    location: 4,
-    frequency: 5,
-    datalinks: 6,
-    documentation: 7
+    topic: 3,
```
benthorner: Could we do a quick change to make this an ordered array and lose the page numbers?

spec/features/create_dataset_spec.rb
benthorner: OK. I don't think anyone would mind, though.

spec/support/dataset_helper.rb
andrewgarner: Done

PR #564, opened 2018-04-18
Remove redundant variables from Heroku schema

benthorner:
The Heroku deployments have their environment set to staging, so we
    (only) need to populate the VCAP_SERVICES variable to provide all of the
    environment variables.
    
    This is not ideal, and we should look at cleaning up the Heroku
    environment variables, as well as improving the setup to include a
    throwaway elasticsearch.
    
    But this is just a quick fix to unblock the signon work.

### DIFF
--- a/app.json
+++ b/app.json
@@ -4,38 +4,8 @@
     "postdeploy": "rake db:migrate ; rake db:seed"
   },
   "env": {
-    "DEVISE_SECRET_KEY": {
-      "required": true
-    },
-    "ES_HOST": {
-      "required": true
-    },
-    "FIND_URL": {
-      "required": true
-    },
-    "HTTP_PASSWORD": {
-      "required": true
-    },
-    "HTTP_USERNAME": {
-      "required": true
-    },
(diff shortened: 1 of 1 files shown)


app.json
```diff
@@ -4,38 +4,8 @@
...
-    "RAILS_ENV": {
-      "required": true
-    },
-    "RAILS_LOG_TO_STDOUT": {
-      "required": true
-    },
-    "RAILS_SERVE_STATIC_FILES": {
-      "required": true
-    },
-    "SECRET_KEY_BASE": {
```
andrewgarner: Should this be:

```json
"SECRET_KEY_BASE": {
  "description": "Cryptographically secure secret key (typically used to generate a secret for cookie sessions)",
  "generator": "secret"
},
```

benthorner: It's populated as part of the VCAP_SERVICES variable, which does helpfully make it similar to staging but also needs a cleanup. This is blocking me from the signon work, though, and I think it would take to much time to properly refactor it now.

PR #565, opened 2018-04-19
Signon

benthorner:


rossjones:
Would it be possible to rebase this as I think a lot of the deletions are already done on master, it might make the PR a bit smaller.

benthorner:
Sure - I was going to leave Andrew alone to review it, but may as well.

app/controllers/home_controller.rb
```diff
@@ -1,27 +0,0 @@
-class HomeController < ApplicationController
```
andrewgarner: We shouldn't be removing this behaviour.

app/mailers/application_mailer.rb
```diff
@@ -1,4 +0,0 @@
-class ApplicationMailer < ActionMailer::Base
```
andrewgarner: I'm not sure we should remove this here. We no longer need to send emails in relation to user accounts but removing wider support for ActionMailer is something we should consider separately in the context of the whole application, not authentication.

app/views/home/dashboard.html.erb
```diff
@@ -1,71 +0,0 @@
-<% content_for :page_title do %>
-  What's available in Publish
```
andrewgarner: We shouldn't just remove this feature. It exists for a reason, we should still be able to provide an overview of what's available in Publish when authenticating through signon.

app/views/home/quality.html.erb
```diff
@@ -1,34 +0,0 @@
-<% content_for :page_title do %>
-    Organisation quality -
```
andrewgarner: We shouldn't remove this.

config/initializers/active_admin.rb
```diff
@@ -1,293 +0,0 @@
-ActiveAdmin.setup do |config|
```
andrewgarner: We should only remove instances where ActiveAdmin has been used to for authentication or the management of users. Removing ActiveAdmin completely from the codebase is something we should review in a wider context.

spec/controllers/datasets_controller_spec.rb
```diff
@@ -1,68 +0,0 @@
-require 'rails_helper'
-
-describe DatasetsController, type: :controller do
```
andrewgarner: This behaviour being tested here is unrelated to signon. We should continue to test it.

spec/features/create_dataset_spec.rb
```diff
@@ -7,11 +7,11 @@
 
   context "when the user goes through entire flow" do
     before(:each) do
-      sign_in_user
+      sign_in_as(user)
     end
 
     it "navigates to new dataset form" do
-      expect(page).to have_current_path("/tasks")
```
andrewgarner: The tasks behaviour should be unaffected by using signon.

spec/models/link_spec.rb
```diff
@@ -1,31 +1,15 @@
 require 'rails_helper'
 
-describe Link, type: :model do
```
andrewgarner: This is unrelated.

spec/models/user_spec.rb
```diff
@@ -1,39 +0,0 @@
-require 'rails_helper'
-
-describe User do
```
andrewgarner: How about we use the `it_behaves_like "a gds-sso user class"` example - https://github.com/alphagov/gds-sso#testing-in-your-application?

spec/rails_helper.rb
```diff
@@ -63,7 +63,3 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
-
-def last_updated_dataset
```
andrewgarner: It would be preferable to propose changes like this in a separate pull-request.

app/controllers/home_controller.rb
benthorner: John was happy with this one. Should be OK.

app/mailers/application_mailer.rb
benthorner: I'm not saying we can never send email, just that we don't need to right now. We can always add it properly (or from the history) if and when we need to.

config/initializers/active_admin.rb
benthorner: I'm hoping we can get to a clean slate - leaving things like this lying around when they can't be used will create bloat.

spec/controllers/datasets_controller_spec.rb
benthorner: This controller test was testing authorisation logic, which I've removed as it's incompatible with the new user model (also half-baked). The tests needed to go as the logic they tested is also gone.

spec/features/create_dataset_spec.rb
benthorner: Unfortunately the home page and related pages are no longer relevant, which is consistent with other GOV.UK admin apps. Jon's happy for now that we redirect to the tasks page, which is consistent with the current behaviour when signing-in.

spec/rails_helper.rb
benthorner: I agree - I intended to do it as part of my let! PR but forgot it. I will try to remember better in future.

app/mailers/application_mailer.rb
rossjones: Please leave it in now.

app/controllers/home_controller.rb
andrewgarner: I've asked @nicholsj about this. The dashboard is a quick way for admins to get an overview but will need to be looked at in the future.

app/models/ability.rb
```diff
@@ -1,11 +0,0 @@
-class Ability
-  include CanCan::Ability
-
-  def initialize(user)
-    user ||= User.new
-
-    can [:read, :update], Dataset do |dataset|
-      user.creator_of_dataset?(dataset) || user.in_organisation?(dataset.organisation)
```
andrewgarner: I've asked @nicholsj for a steer on this. He said that we definitely need permissions for editing (and creating) datasets and that they should be tied to the owner organisation of the dataset.

app/views/home/dashboard.html.erb
nicholsj: Sorry to jump in  but hope it's useful. I'm not sure I did know this would be removing the "What's in Publish" dashboard (sorry if I gave that impression at some point). While that ashboard obvs needs some discussion/work, I think we should keep it for now. I can't see how it's related to SignOn?

app/views/home/quality.html.erb
nicholsj: Similar to my comment on the dashboard - let's keep this. We'll have to work out access controls for some of these pages, which I assume is why it's come up now, but that can come later

app/views/home/index.html.erb
```diff
@@ -1,35 +0,0 @@
...
-      </h1>
-      <p class="lede">
-        You must create an account to use this service.
-      </p>
-      <p class="lede">
-        If you already have an account, sign in to publish a new dataset
-        or update an existing one.
-      </p>
-
-      <a class="button dgu-masthead__button-get-started" href="<%= new_user_session_path %>" role="button">
```
andrewgarner: I've asked @nicholsj about this one too. His steer was that the public facing page should remain and the signon part for now should be limited to changing the "Sign in" button to use the authentication flow provided by Signon instead of our own UI.

PR #566, opened 2018-04-19
Remove weird things

benthorner:


app/controllers/datasets/datafiles_controller.rb
```diff
@@ -1,4 +1,3 @@
-# coding: utf-8
```
andrewgarner: This is included because the file has UTF-8 characters. For example:

```
flash[:alert] = "Are you sure you want to delete ‘#{@datafile.name}’?"
```

benthorner: Ah, sorry - I assumed it was something someone's editor inserted.

andrewgarner: I think that's a correct assumption, but it would have been because of those quote characters.

benthorner:
Yup - updated. I feel like whoever put it in there in the first place should have the joy of removing it :-(

PR #567, opened 2018-04-19
Move UrlValidator to its proper place

benthorner:


spec/spec_helper.rb
```diff
@@ -28,6 +28,10 @@
     allow_any_instance_of(UrlValidator).to receive(:validPath?).and_return(true)
   end
 
+  config.before(:each) do
```
andrewgarner: This is duplication of lines 27:29.

benthorner:
Putting an app dependency in lib was causing me a lot of grief trying to run the tests, because those files aren't loaded properly. I would very much like to put it in the app folder.

benthorner:
Maybe in a new validations subfolder? I've seen that before as well.

benthorner:
Also, turns out they are a type of concern: https://github.com/rails/rails/blob/master/activemodel/lib/active_model/validations.rb.

andrewgarner:
I think you can solve that by adding `config.autoload_paths += %w(#{config.root}/lib)` to `application.rb`.

spec/spec_helper.rb
benthorner: Sorry - bad rebase.

benthorner:
I'm aware you can do that, but models/concerns is that right place and we're shooting ourselves in the foot by going against rails conventions.

andrewgarner:
> Also, turns out they are a type of concern: https://github.com/rails/rails/blob/master/activemodel/lib/active_model/validations.rb.

We've not modelled `UrlValidator` as a concern though. The example you gave extends [ActiveSupport::Concern](http://api.rubyonrails.org/v5.1/classes/ActiveSupport/Concern.html) and implements `included` and `class_methods` allowing it be used like this:

```rb
class Comment
  include ActiveModel::Validations

  validate do |comment|
    comment.must_be_friends
  end

  def must_be_friends
    errors.add(:base, 'Must be friends to leave a comment') unless commenter.friend_of?(commentee)
  end
end
```

rossjones:
Just to add another opinion, `app/validators/` .. 

Maybe given lack of concensus we should just leave it where it is for now?

benthorner:
So I think this is about more than this one file i.e. we're looking to set a convention. I think the discussion is about whether we keep all our app code in app/, or put some of it in lib/.

Andrew's thoughts on this are that the code in lib/ should be code that is separable from our normal domain logic - it's reusable. This applies to UrlValidator because it doesn't refer to any app models.

I'm used to seeing all of the code in app/ unless it's for a Rake task. The benefit of putting everything in app/ is rails will auto-require it; currently running a single test breaks because of brittle requires.

rossjones:
From what I've read, lib/ being for non-domain specific stuff seems to be the norm.  
Looks like https://github.com/alphagov/whitehall/tree/master/lib uses lib/ quite a lot for that approach, so we should follow their convention rather than setting a new one.

benthorner:
OK. I'll replace this PR with one to autoload lib/validators and thereby fix the issue with running a single test.

PR #568, opened 2018-04-19
Speed up import by not asking for the same info on each dataset

rossjones:
Recent refactor means that the organisation and topic caches were
being refreshed on each run which is spectacularly inefficient.


rossjones:
@andrewgarner Could you review this and I'll use it on prod as it's about 20 times faster...

PR #569, opened 2018-04-20
Use let!(:var) instead of let(:var); var

benthorner:


spec/features/edit_dataset_spec.rb
```diff
@@ -19,8 +19,6 @@
                                                  owner: user ) }
 
   before(:each) do
-    allow_any_instance_of(UrlValidator).to receive(:validPath?).and_return(true)
```
andrewgarner: This seems unrelated to your commit?

benthorner: It's duplicated from the spec_helper.

benthorner: I don't think anyone's going to miss it ;-)

PR #570, opened 2018-04-20
Cleanup files

benthorner:
This PR is an attempt to cleanup some of the accumulated cruft in Publish.

   - Centralise the basic README / setup docs
   - Remove redundant files and plugins


README.md
```diff
@@ -1,115 +1,38 @@
...
-## Fake user accounts
-To log in as a fake user, use the credentials in 'seeds.rb' in the 'db' folder.
+# import production data - ctrl-c after ~1000 datasets
+bin/import
 
-## Importing data
-You can import data into Postgres using following commands:
+# start a web server
+rails s
+`
```
rossjones: Stray ` ?

README.md
```diff
@@ -1,115 +1,38 @@
...
-rails spec
-```
+`
+# install dependencies
+bin/setup
 
-## Fake user accounts
-To log in as a fake user, use the credentials in 'seeds.rb' in the 'db' folder.
+# import production data - ctrl-c after ~1000 datasets
+bin/import
```
rossjones: Might it also be worth mentioning `scripts/replication.sh`?

README.md
```diff
@@ -1,115 +1,38 @@
...
-# Vagrant
-
-To run the app in a local VM with vagrant, install Vagrant and Virtualbox, then:
-```
-$ vagrant up
-$ vagrant ssh -c /vagrant/tools/vagrant-dev-setup.sh
-$ vagrant ssh -c "cd /vagrant && rails s"
-$ vagrant ssh -c "cd /vagrant && bundle exec sidekiq
-```
+See [here](doc/architecture/README.md) for all of our Architecture Decision Records.
```
rossjones: They're actually in `docs/arch`

package.json
```diff
@@ -1,11 +0,0 @@
-{
-  "name": "publish_data_beta",
-  "private": true,
-  "dependencies": {},
-  "devDependencies": {
-    "eslint": "^4.0.0",
```
rossjones: Do we no longer need/use eslint?  It seems like a useful thing to use...

README.md
benthorner: Actually should been ``` - forgot my MD :-)

README.md
benthorner: I moved them as part of this PR so they're the same as govuk-aws (just so we're that bit more consistent).

README.md
benthorner: It didn't realise this was here - does this replace bin/import + lib/tasks/import.rake? It seems a lot simpler and should achieve the same effect.

rossjones: It should for us, but it won't for anyone outside of our team trying to use it.

package.json
benthorner: It wasn't being used previously. Perhaps a separate PR like the one for govuk-lint (or even better, try and get rid of the JS code ;-)

README.md
benthorner: OK. I think we should try and minimise the number of setup paths we've got and make them robust. I've not found a need for replication.sh - could we lose that one?

I'm also wondering whether we could download a small version of the other CSVs and combine the import.rake with db/seeds.rb so it's all in one obvious place.

But that's also another PR I reckon!

andrewgarner: The `replication.sh` script allows us to take a copy of the imported data so you can run the application locally with representative data. It quicker and easier than having to manually import and process the data in the CKAN dump files to achieve the same result.

Additionally, it will also replicate datasets that have been manually published or edited, the `bin/import` doesn't do that.

andrewgarner: We should also consult the git history for context about how certain features and behaviour came to be.

benthorner: It looks like Laurent and Max wrote the setup.sh file, but I bet they didn't know about this script. replication.sh is much faster, requires less code and gives me all the seed data I need.

Also, if the API dump isn't going to continue in the future, this method is going to be more robust. In an effort to simplify, can we remove setup.sh + associated rake tasks?

PR #574, opened 2018-04-20
Fix for missing default rake task

andrewgarner:
To fix deploy failures:

```
rake aborted!
Don't know how to build task 'default' (see --tasks)
```

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,4 @@ require_relative 'config/application'
 
 Rails.application.load_tasks
 
-task default: [:spec]
+task default: %i(spec lint)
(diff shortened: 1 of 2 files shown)


benthorner:
I'm getting '0 files inspected, no offenses detected' when I run rake lint, even when I modify a file.

benthorner:
I modified app/admin/.... I would expect it to inspect the file. I also can't see the --diff option in the CLI doc.

andrewgarner:
Unfortunately you have to go and read the source code to understand the behaviour – https://github.com/alphagov/govuk-lint/blob/master/lib/govuk/lint/diff.rb

Essentially the diff is files you have changed and committed.

benthorner:
I have to change it to 'govuk-lint-ruby --cache=true --format clang' to get it to do anything, which then prints out that we've got 4 new violations in delete.rake on master.

PR #577, opened 2018-04-23
Remove active admin

benthorner:
Activeadmin provides a backend admin management workflow that isn't currently used. Let's remove it until it's clear what we actually need.

https://trello.com/c/24SchQBN/76-remove-activeadmin


andrewgarner:
Is there a Trello card for this?

PR #582, opened 2018-04-24
Remove unused active?/closed? org methods

benthorner:


spec/models/organisation_spec.rb
```diff
@@ -1,16 +1,6 @@
 require 'rails_helper'
 
 describe Organisation do
-  it "can create a new dataset" do
```
emmabeynon: It feels like it's useful to test that a new dataset can be created, so I don't think we should delete the whole test, but rather remove lines 10 and 11 which test the deleted methods.

benthorner: So the reasoning here is that this is an ActiveRecord model, which means we can take certain abilities (like creation) for granted. We're effectively trying to test other people's code, which we should avoid - one clue is that none of the other models have a test like this ;-)

emmabeynon: Thinking about it a bit more, creating an organisation seems to be covered by the other tests anyway, so probably good to delete this one after all 👍

PR #585, opened 2018-04-25
Remove last remains of active admin

benthorner:
This is causing a bug on Heroku as these stray files can't be loaded
properly.


benthorner:
Good spot - I've done it for this and the previous one. We should probably do that this afternoon so this doesn't happen again.

PR #590, opened 2018-05-01
Add datafile_last_updated_at attribute to dataset

emmabeynon:

There are instances in Find Data where there is a mismatch between a dataset's "Last updated" timestamp that appears on the search results page and the timestamp that appears on the dataset's own page.  We've agreed that the "Last updated" date should reflect the last time a datafile was modified, and if there are no datafiles present then the last modified metadata date should be used.  In preparation for this we:

1) Add a `datafile_last_updated_at` timestamp to the Dataset model so we can track this more accurately in future.
2) Modify the dataset importer to set `datafile_last_updated_at` to the time that the most recently modified datafile was updated.  In cases where no datafiles have been updated, we use the most recently created datafile's created at date.
3) Modify the dataset importer to set each datafile's `updated_at` to its imported datafile's `last_modified_at` value if present, else its `created` value.

A separate PR will be opened to backfill `datafile_last_updated_at` dates for all datasets and a separate card will be written up to deal with setting a value for this attribute when a datafile is modified in Publish.

Related PRs: #591 & #592

### DIFF
--- a/app/services/legacy/dataset_import_service.rb
+++ b/app/services/legacy/dataset_import_service.rb
@@ -51,7 +51,8 @@ class Legacy::DatasetImportService
       licence_custom: get_extra("licence"),
       topic_id: build_topic_id,
       secondary_topic_id: build_secondary_topic_id,
-      status: "published"
+      status: "published",
+      datafile_last_updated_at: most_recently_updated_datafile_date,
     }
   end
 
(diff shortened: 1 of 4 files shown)


app/services/legacy/dataset_import_service.rb
```diff
@@ -103,7 +104,7 @@ def base_attributes_for_resource(resource)
       format: resource["format"],
       name: datafile_name(resource),
       created_at: resource["created"] || dataset.created_at,
-      updated_at: dataset.last_updated_at
+      updated_at: resource["last_modified_at"] || resource["created"]
```
andrewgarner: Should this be `resource["last_modified_at"] || resource["created"] || dataset.created_at`?

There looks to be a case where `resource["created"]` can be nil given the way `created_at` is set.

emmabeynon:
Closed in favour of https://github.com/alphagov/datagovuk_publish/pull/602

PR #591, opened 2018-05-01
Add metadata_last_updated_at attribute to Dataset

emmabeynon:
Note: this is branched off https://github.com/alphagov/datagovuk_publish/pull/590.  After it has been merged, this PR will be rebased.

There are instances in Find Data where there is a mismatch between a dataset's "Last updated" timestamp that appears on the search results page and the timestamp that appears on the dataset's own page. We've agreed that the "Last updated" date should reflect the last time a datafile was modified, and if there are no datafiles present then the last modified metadata date should be used. In preparation for this we:

1. Add a `metadata_last_updated_at` timestamp to the Dataset model so we can track this more accurately in future.
2. Modify the dataset importer to set `metadata_last_updated_at` to the `metadata_modified` value provided by the legacy dataset.

A separate PR will be opened to backfill `metadata_last_updated_at` dates for all datasets and a separate card will be written up to deal with setting a value for this attribute when metadata is modified in Publish.

Related PRs: #590 & #592

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -88,6 +88,7 @@ class Dataset < ApplicationRecord
                 created_at
                 harvested
                 uuid
+                datafile_last_updated_at
             ],
       include: {
         organisation: {},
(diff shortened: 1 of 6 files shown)


emmabeynon:
Closed in favour of #602

PR #592, opened 2018-05-01
Change last_updated_at on Dataset to public_updated_at

emmabeynon:
Note: this is branched off #591. After it has been merged, this PR will be rebased.

There are instances in Find Data where there is a mismatch between a dataset's "Last updated" timestamp that appears on the search results page and the timestamp that appears on the dataset's own page. We've agreed that the "Last updated" date should reflect the last time a datafile was modified, and if there are no datafiles present then the last modified metadata date should be used. In preparation for this we:

1. Rename `last_updated_at` on Dataset to `public_updated_at` to reflect that this is a public facing date which will be shown in search results against "Last updated".
2. Update the Dataset Importer service to set `public_updated_at` to the same value as `datafile_last_updated_at`, or in cases where datafiles aren't present set it to the same value as `metdata_last_updated_at`.

A separate PR will deal with backfilling `public_updated_at` and a card will be written to deal with correctly setting `public_updated_at` in Publish.

Related PRs: #590 & #591

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -88,6 +88,7 @@ class Dataset < ApplicationRecord
                 created_at
                 harvested
                 uuid
+                datafile_last_updated_at
             ],
       include: {
         organisation: {},
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -191,6 +191,8 @@ private
   end
 
   def set_last_updated_date
-    self.last_updated_at = Time.now
+    #TO DO: this needs to be set to the most recently updated datafile update date
+    # or the last modified metadata date
+    self.public_updated_at = Time.now
   end
 end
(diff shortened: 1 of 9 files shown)


app/models/dataset.rb
```diff
@@ -190,6 +191,8 @@ def set_first_publication_date
   end
 
   def set_last_updated_date
```
andrewgarner: Should we rename this to `set_public_updated_at`?

app/services/legacy/dataset_import_service.rb
```diff
@@ -32,7 +32,7 @@ def dataset_attributes
       frequency: build_frequency,
       published_date: legacy_dataset["metadata_created"],
       created_at: legacy_dataset["metadata_created"],
-      last_updated_at: legacy_dataset["metadata_modified"],
```
andrewgarner: The Find application is dependent on `last_updated_at` being present:

https://github.com/alphagov/datagovuk_find/blob/201a714970e1b35c78960606588de1bb6c1b6641/app/services/search/query.rb#L182

Therefore we need to make these changes in a backward compatible way so that we don't require any downtime.

1. Add `public_updated_at` field, with `last_updated_at` being an alias. Release and re-index.
2. Release a new version of find that now uses `public_updated_at`.
3. Remove `last_updated_at` from Publish,  release and re-index.

emmabeynon:
Closed in favour of #602

PR #595, opened 2018-05-02
Move legacy API interaction into a service

rossjones:
Removes all knowledge of Legacy API from the import rake task, and
instead handles it in a service class that exposes methods
to call the API (dataset_show and publisher_show to mimic the api
function names).


rossjones:
Yeah I was going to ask where it should go. There's a gem that does most of the lifting, so I may eventually go with that - for now, will move it (and tests)

benthorner:
Could it be called Legacy::APIService as Legacy::LegacyAPIService seems a bit overkill ;-)

PR #598, opened 2018-05-02
Fix alias updater in dev and make it atomic

benthorner:
https://trello.com/c/KsehMWJo/162-fix-reindexing-not-working-in-dev

This has been making https://trello.com/c/wQjbYbbv/50-last-updated-different-in-search-results-vs-dataset-page more difficult to complete.

The alias updater required an environment variable to specify the
current ES index to replace, which was really annoying to get manually.
Now the current index is determined automatically.

The alias updater used to remove the old alias link and serially and the
new one. A failure at this point would have caused Find to stop working.
Now the alias update (remove old link, add new link) is done atomically.

### DIFF
--- a/app/services/alias_updater_service.rb
+++ b/app/services/alias_updater_service.rb
@@ -7,45 +7,35 @@ class AliasUpdaterService
   end
 
   def run
-    remove_alias_from_old_index
     assign_alias_to_new_index
-    logger.info "Alias '#{index_alias}' now pointing to '#{new_index_name}'"
+    @logger.info "Alias '#{@index_alias}' -> '#{@new_index_name}'"
   end
 
 private
 
-  attr_reader :logger, :client, :index_alias, :new_index_name
-
-  def remove_alias_from_old_index
-    client.indices.update_aliases body: {
-      actions: [
-        {
-          remove: {
-            index: Dataset.index_name,
-            alias: index_alias
-          }
-        }
-      ]
-    }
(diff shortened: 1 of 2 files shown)


app/services/alias_updater_service.rb
```diff
@@ -7,45 +7,35 @@ def initialize(args)
...
   def run
-    remove_alias_from_old_index
     assign_alias_to_new_index
-    logger.info "Alias '#{index_alias}' now pointing to '#{new_index_name}'"
+    @logger.info "Alias '#{@index_alias}' -> '#{@new_index_name}'"
   end
 
 private
 
-  attr_reader :logger, :client, :index_alias, :new_index_name
```
andrewgarner: Why do we need to remove the accessor methods to enable this behaviour?

benthorner: Mine and Emma's reaction when reading the code was "where do these mysterious variables come from". I think using instance variables requires less code and makes it obvious where they're coming from.

benthorner: This should be OK - me and Andrew had a chat on the phone :-)

rossjones:
How do we test this?

benthorner:
I did add a unit test. Emma is going to test locally, and we can do a run on staging as well. Would that be suitable?

benthorner:
Oops - spotted your comment too late :-|

PR #599, opened 2018-05-02
Choose the correct name for data files

rossjones:
When importing a legacy datafile, the importer was previously just using
the description field.  It should have been checking resource['name']
and only using resource['description'] as a fallback. This resulted in
lots of datasets having links called 'Resource locator' which is a
description of the link, rather than the name.

https://trello.com/c/b9VO8lOm/71-when-syncing-importing-we-end-up-with-links-that-have-no-name-when-they-should

### DIFF
--- a/app/services/legacy/dataset_import_service.rb
+++ b/app/services/legacy/dataset_import_service.rb
@@ -127,8 +127,8 @@ class Legacy::DatasetImportService
   end
 
   def datafile_name(resource)
-    name = resource.fetch('name', '').strip
-    name = resource.fetch('description', '').strip if name.blank?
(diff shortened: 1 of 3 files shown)


app/services/legacy/dataset_import_service.rb
```diff
@@ -127,7 +127,9 @@ def parse_date(date)
   end
 
   def datafile_name(resource)
-    resource['description'].strip == '' ? 'No name specified' : resource['description']
+    name = resource.fetch('name', '').strip
```
andrewgarner: http://api.rubyonrails.org/classes/String.html#method-i-squish might be a good option in this instance

PR #600, opened 2018-05-02
Remove unused tests for now

benthorner:
We should invest some time to create proper e2e tests in future, but
this approach is unlikely to work with Signon and is currently only
serving to hide the lack of testing. I think we should remove these
tests to make it clear that some new e2e tests are required.


rossjones:
@benthorner cf. https://github.com/alphagov/datagovuk-end-to-end-tests

benthorner:
@rossjones One day ;-)

PR #602, opened 2018-05-02
Send public_updated_at to elasticsearch

emmabeynon:

There are instances in Find Data where there is a mismatch between a dataset's "Last updated" timestamp that appears on the search results page and the timestamp that appears on the dataset's own page. We've agreed that the "Last updated" date should reflect the last time a datafile was modified, and if there are no datafiles present then the date that metadata was last modified should be used. In this PR we present a `public_updated_at` value to the Elasticsearch index which Find Data can use to display a "Last updated" date which satisfies the above logic.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -89,6 +89,7 @@ class Dataset < ApplicationRecord
                 harvested
                 uuid
             ],
+      methods: :public_updated_at,
       include: {
         organisation: {},
         topic: {},
@@ -168,6 +169,10 @@ class Dataset < ApplicationRecord
     %w[annually quarterly monthly].include?(frequency)
   end
 
+  def public_updated_at
+    most_recently_updated_datafile_timestamp || self.updated_at
+  end
+
 private
 
   def send_to_search_index
@@ -191,5 +196,10 @@ private
 
   def set_last_updated_date
     self.last_updated_at = Time.now
+  end
+
+  def most_recently_updated_datafile_timestamp
+    timestamps = self.datafiles.map(&:updated_at)
+    timestamps.sort.last if timestamps.present?
   end
 end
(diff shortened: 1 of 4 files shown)


spec/services/legacy/dataset_import_service_spec.rb
```diff
@@ -53,18 +54,20 @@
     end
 
     it "creates the datafiles for the imported dataset" do
+      first_resource = legacy_dataset["resources"][0]
+      first_resource["last_modified_at"] = "2017-12-14T09:35:25.928982"
```
benthorner: Could we use `Time.now` here instead of a constant. I usually interpret a constant as something special, but I think this is just a throwaway value.

emmabeynon: Sure, I think that might have been a hangover from a previous implementation.

andrewgarner:
Based on https://github.com/alphagov/datagovuk_find/pull/597 we'll need to first release a find PR that adds support for `public_updated_at`

app/models/dataset.rb
```diff
@@ -168,6 +169,10 @@ def timeseries?
     %w[annually quarterly monthly].include?(frequency)
   end
 
+  def public_updated_at
+    most_recently_updated_datafile_timestamp || self.updated_at
```
rossjones: I might be missing some context, should this be the most recent of the two dates? Or is public_updated_at solely about the update time of the data files, and the metadata record updated_at is just a backup?

rossjones: Okay. I was just concerned about legacy import, but that should be okay as the updated_at should mostly be the same.  Thanks for the clarification.

spec/services/legacy/dataset_import_service_spec.rb
emmabeynon: Seems I forgot to push! Now done 👍

PR #605, opened 2018-05-03
Add extra licence options

maxgds:
https://trello.com/c/XRQKhz80/158-add-additional-licence-fields

This adds the extra licence types, and also pulls the options into a partial so it can be shared between the create and edit flows.

### DIFF
--- a/app/views/datasets/licences/_licence_options.html.erb
+++ b/app/views/datasets/licences/_licence_options.html.erb
@@ -1,0 +1,49 @@
+<div class="form-group">
+  <fieldset>
+    <legend class="visually-hidden">Choose a licence for this dataset</legend>
+
+    <%= dataset_field form, @dataset,
+    name: 'licence',
+    input_type: :radio_button,
+    label: 'Open Government Licence',
+    value: 'uk-ogl' %>
+
+    <%= dataset_field form, @dataset,
+    name: 'licence',
+    input_type: :radio_button,
+    label: 'Open Data Commons Attribution License',
+    value: 'odc-by' %>
+
+    <%= dataset_field form, @dataset,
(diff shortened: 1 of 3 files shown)


app/views/datasets/licences/edit.html.erb
```diff
@@ -23,32 +23,7 @@
...
-        <div class="panel panel-border-narrow js-hidden" id="other-licence">
-          <%= dataset_field f, @dataset,
-          input_type: :text_field,
-          name: 'licence_other',
-          label: 'Name of your licence:' %>
-        </div>
-      </fieldset>
-    </div>
-
+    <%= render partial: 'licence_options', :locals => {f: f} %>
```
andrewgarner: It's good practice to _tell_ the partial everything it needs to know to render to make it more reusable.

We should tell the partial about the dataset so that it doesn't depend on `@dataset` being available in it's rendering context.

```rb
<%= render partial: 'licence_options', locals: { f: f, dataset: @dataset } %>
```

app/views/datasets/licences/_licence_options.html.erb
```diff
@@ -0,0 +1,49 @@
+<div class="form-group">
+  <fieldset>
+    <legend class="visually-hidden">Choose a licence for this dataset</legend>
+
+    <%= dataset_field f, @dataset,
```
andrewgarner: Perhaps we should set the local variable to `form` not `f` as `f` is ambiguous now we don't have the `<%= form_for @dataset, url: update_dataset_licence_path(@dataset.uuid, @dataset.name) do |f| %>` context.

maxgds:
@andrewgarner I made the recommended changes. Good to go?

PR #607, opened 2018-05-03
Make sure Heroku runs migrations every deploy

benthorner:
Unfortunately the postdeploy script in app.json only runs when the app is first created. There's apparently no hook for something that runs every time it's deployed, apart from the start command.

### DIFF
--- a/Procfile
+++ b/Procfile
@@ -1,2 +1,3 @@
+release: bin/rails db:migrate
 web: bin/rails server -p $PORT -e $RAILS_ENV
 worker: bin/sidekiq


benthorner:
Ah, I didn't expect they would solve it there. Let's try it.

Procfile
```diff
@@ -1,2 +1,3 @@
+release: rake db:migrate
```
andrewgarner: I think we should use the bin stub to be consistent with the other processes. Also `bin/rails` is the [new default for running commands](http://guides.rubyonrails.org/command_line.html#bin-rails).

```
release: bin/rails db:migrate
```

benthorner: I was being consistent with app.json, but OK. I wonder how long this 'default' will take to become the default ;-)

andrewgarner: I'm fighting the good fight ;-)

PR #608, opened 2018-05-03
Add org mappings for user study

benthorner:


benthorner:
I wasn't quite ready ;-) Try now.

PR #610, opened 2018-05-03
Update view to show additional licence types

maxgds:
Fix for the blank licence type bug.

We should probably store the types in a config file somewhere and loop through them where we need to display, but this quick fix plugs the gap to get us ready for testing.

### DIFF
--- a/app/views/datasets/show.html.erb
+++ b/app/views/datasets/show.html.erb
@@ -68,7 +68,21 @@
             <span class="error-message error">You must choose a licence</span>
           </div>
         <% end %>
-        <%= @dataset.licence == 'uk-ogl' ? 'Open Government Licence' : @dataset.licence_other %>
+        <%= case @dataset.licence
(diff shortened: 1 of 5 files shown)


app/views/datasets/show.html.erb
```diff
@@ -68,7 +68,21 @@
...
+          when 'odc-odbl'
+            'Open Data Commons Open Database License (ODbL)'
+          when 'cc-by'
+            'Creative Commons Attribution'
+          when 'cc-zero'
+            'Creative Commons CCZero'
+          else
+            'Other'
+          end
+        =%>
```
andrewgarner: There is a stray `=`. It should be `%>`

maxgds:
@andrewgarner I fixed the stray =

andrewgarner:
Closing this so @maxgds can raise a new PR to fix the broken rebase.

PR #612, opened 2018-05-03
Update view to show additional licence types

maxgds:
https://trello.com/c/XRQKhz80/158-add-additional-licence-fields

Fix for missing licences in view at end of creating new dataset.

(Replaces corrupted PR #610)


benthorner:
I think we could use Rails locales instead of a case statement here. As we would really like to get this finished for the user study, I'm going to try and get this done.

PR #613, opened 2018-05-04
Make sure ES index exists in Heroku

benthorner:


Procfile
```diff
@@ -1,3 +1,3 @@
-release: bin/rails db:migrate
+release: bin/rails db:migrate && bin/rails search:reindex
```
andrewgarner: `bin/rails` accepts multiple tasks so you can do `bin/rails db:migrate search:reindex`

benthorner: True.

PR #620, opened 2018-05-14
Refactor dataset error banners into partial

benthorner:
https://trello.com/c/VgPz1E2A/170-refactor-error-banner-into-a-shared-partial-file


benthorner:
:-)

PR #621, opened 2018-05-14
Remove redundant div

benthorner:
https://trello.com/c/VgPz1E2A/170-refactor-error-banner-into-a-shared-partial-file


benthorner:
I agree. Git doesn't like changing indentation!

PR #625, opened 2018-05-15
Add worker to export internal model metrics

benthorner:
https://trello.com/c/HsPIFLVB/187-expose-internal-metrics-using-prometheus-grafana


spec/rails_helper.rb
```diff
@@ -8,60 +8,28 @@
 require 'rspec/rails'
 require 'capybara/rspec'
 require 'capybara/rails'
-# Add additional requires below this line. Rails is not loaded until this point!
```
andrewgarner: Can we keep this inline with the generator output as per https://github.com/alphagov/datagovuk_find/pull/608?

spec/rails_helper.rb
```diff
@@ -8,60 +8,28 @@
...
-# directory. Alternatively, in the individual `*_spec.rb` files, manually
-# require only the support files necessary.
 #
 Dir[Rails.root.join('spec/support/**/*.rb')].each { |f| require f }
 
-# Checks for pending migration and applies them before tests are run.
-# If you are not using ActiveRecord, you can remove this line.
 ActiveRecord::Migration.maintain_test_schema!
 
+GDS::Metrics.configure do |config|
```
andrewgarner: Can this be moved to `spec/features/model_metrics_spec.rb` or do we need this setup before every spec?

benthorner: I think it should be in rails_helper as it's part of the environment setup (it doesn't vary with tests). Otherwise, the test environment would change depending on when the associated test file is run.

spec/rails_helper.rb
benthorner: Done - I was actually going to do another PR to replicate what you did in Find. But I don't think we have time now.

benthorner:
Unfortunately I didn't consider that we're running Sidekiq in a separate container, which means the cached metrics aren't accessible to the rails container, which is what exposes /metrics.

I've solved this by replacing Sidekiq scheduler with Rufus scheduler running inside rails, but I think this is now a bit bespoke. I still think this metric data is valuable, and we will certainly use the Postgres exporter when it comes out, so this is only a temporary solution. Do we still want to go ahead with this, given the scheduling is now a bit 'surprising'.

andrewgarner:
I'd prefer to wait until the Postgres exporter is generally available.

PR #628, opened 2018-05-16
Remove redundant active column for organisations

benthorner:
This column isn't used anywhere and it's only populated manually by a
rake task, making it unreliable. If we need this column, we should add
it in such a way that it is populated reliably.


benthorner:
Sorry, corrected - too many branches. The active column is populated by looking at another column in the same table (replace_by). It's not used anywhere and it would be better to calculate it on-the-fly if we actually did need it somewhere. This PR is just to clear the way.

rossjones:
My concern is, that once we stop doing imports of organisations from legacy, we'll have no way of knowing which organisations are active and which aren't.  Some closed organisations are replaced by others, but some aren't.  

If we're not using the active field to show and explain about closed organisations, then that's probably a bug as much as a candidate for removal.

benthorner:
That's definitely a valid concern, but it would still be something to implement when we add the ability to close an organisation. Currently, there's no such feature and all the models have 'active = nil'.

PR #629, opened 2018-05-16
Rake task to reset all of the datasets for a single organisation

rossjones:
As we encounter issues with the synchronisation, we occassionally get
publishers running their harvester and ending up with 1400 datasets that
have not synced to publish, and fall outside the window within which we
check for changes.

Running this task, with the short name of an organisation, and a boolean
will:

* Delete all the datasets for the organisation if the boolean is "true"
* Re-import all of the datasets for the organisation from the legacy API

There is no transaction, if the container runs out of memory half way
through, you need to run it again! This should only be used in dire
emergencies.

### DIFF
--- a/lib/tasks/import.rake
+++ b/lib/tasks/import.rake
@@ -79,6 +79,45 @@ namespace :import do
       indexer.index(dataset.uuid)
     }
   end
+
+  desc "Resets all of an organisation's datasets by (optionally) removing the existing ones and rebuilding from legacy API"
+  task :organisation_datasets, %i[legacy_shortname delete] => :environment do |_, args|
+    organisation = Organisation.find_by(name: args.legacy_shortname)
+    if organisation.nil?
+      Rails.logger.error "Unable to find organisation '#{args.legacy_shortname}''"
+      next
+    end
+
+    indexer = Legacy::DatasetIndexService.new
+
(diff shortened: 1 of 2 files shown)


lib/tasks/import.rake
```diff
@@ -79,6 +79,45 @@ namespace :import do
...
+    total = search.call(0, 0)
+
+    (0..total['count']).step(1000) do |offset|
+      response = search.call(1000, offset)
+      datasets = response['results']
+
+      datasets.each { |dataset|
+        Legacy::DatasetImportService.new(dataset, organisation_cache, topic_cache).run
+        indexer.remove_from_index(dataset['id'])
+        indexer.index(dataset['id'])
```
andrewgarner: Did you confirm we need these two steps? I wonder if they happen automatically through callbacks on the dataset?

rossjones: They definitely didn't happen when I did import:single_legacy_dataset - I ended up with duplicates.  I think it would be worth more investigation, because I think an update_index would be more useful than deleting and re-adding generally.  

Just not sure that investigation/refactor is small enough to fit in this PR.

andrewgarner: No worries. Hopefully it's only a temporary measure too.

PR #630, opened 2018-05-17
Replace PLEK URI override with GOVUK app domain

benthorner:
https://trello.com/b/a30DSRUe/dgu-q1-doing

Rather than specifying a custom URI for signon, PLEK expects a variable
called GOVUK_APP_DOMAIN to be set, which it will then use to build the
signon URI and any other URIs.

### DIFF
--- a/app.json
+++ b/app.json
@@ -22,8 +22,8 @@
     "OAUTH_SECRET": {
       "required": false
     },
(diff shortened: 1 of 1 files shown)


benthorner:
@andrewgarner You'll need to modify the app secrets e.g. staging will have `GOVUK_APP_DOMAIN=staging.publishing.service.gov.uk`.

PR #634, opened 2018-05-18
Export dataset released status to ES

benthorner:
https://trello.com/c/SkQ9KxWf/139-show-availability-of-datasets-in-search-results

Previously Find would try to infer the dataset released status by using
the count of links. However, the nested link documents are not returned
for certain queries, meaning the released status could not be
determined. Exporting the attribute as a static value solves this
problem.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -61,8 +61,7 @@ class Dataset < ApplicationRecord
   # What we actually want to index in Elastic, rather than the whole
   # dataset.
   def as_indexed_json(_options = {})
-    as_json(
-      only: %i[
+    as_json(only: %i[
                 name
                 legacy_name
                 title
@@ -89,15 +88,17 @@ class Dataset < ApplicationRecord
                 harvested
                 uuid
             ],
-      methods: :public_updated_at,
-      include: {
-        organisation: {},
-        topic: {},
-        datafiles: {},
-        docs: {},
-        inspire_dataset: {}
-      }
-    )
+          methods: %i[
+            public_updated_at
+            released
+          ],
+          include: {
+            organisation: {},
+            topic: {},
+            datafiles: {},
+            docs: {},
+            inspire_dataset: {}
+          })
   end
 
   def owner
@@ -171,6 +172,10 @@ class Dataset < ApplicationRecord
 
   def public_updated_at
     most_recently_updated_datafile_timestamp || self.updated_at
+  end
+
+  def released
+    links.count.positive?
   end
 
 private
--- a/spec/models/dataset_spec.rb
+++ b/spec/models/dataset_spec.rb
@@ -123,4 +123,19 @@ describe Dataset do
       expect(dataset.public_updated_at).to eq(dataset.updated_at)
     end
   end
+
+  describe "#released" do
+    it "returns true when there is a link" do
+      dataset = FactoryGirl.create(:dataset, :with_datafile)
+      expect(dataset.released).to be_truthy
+
+      dataset = FactoryGirl.create(:dataset, :with_doc)
+      expect(dataset.released).to be_truthy
+    end
+
+    it "returns false when there is no link" do
+      dataset = FactoryGirl.create(:dataset)
+      expect(dataset.released).to be_falsey
+    end
+  end
 end
--- a/spec/factories/dataset.rb
+++ b/spec/factories/dataset.rb
@@ -9,5 +9,13 @@ FactoryGirl.define do
     licence "uk-ogl"
     topic
     last_updated_at Time.now
+
+    trait :with_datafile do
+      datafiles { create_list(:datafile, 1) }
+    end
+
+    trait :with_doc do
+      docs { create_list(:doc, 1) }
+    end
   end
 end


app/models/dataset.rb
```diff
@@ -61,43 +61,22 @@ def publish!
   # What we actually want to index in Elastic, rather than the whole
   # dataset.
   def as_indexed_json(_options = {})
-    as_json(
-      only: %i[
-                name
```
andrewgarner: By having one attribute per line we keep the diff to just that line when we add or remove an attribute.

spec/models/dataset_spec.rb
```diff
@@ -123,4 +123,19 @@
       expect(dataset.public_updated_at).to eq(dataset.updated_at)
     end
   end
+
+  describe "#released" do
+    it "returns true when there is a link" do
+      dataset = FactoryGirl.create(:dataset, datafiles: [FactoryGirl.create(:datafile)])
+      expect(dataset.released).to be_truthy
```
andrewgarner: How about we use a predicate matcher?

```rb
expect(dataset).to be_released

expect(dataset).not_to be_released
```

You could use RSpec's `context` to describe the different way `released` behaves:

```rb
context 'when there are datafiles' do
  let(:dataset) { FactoryGirl.create(:dataset, datafiles: [FactoryGirl.create(:datafile)]) }

  it { is_expected.to be_released }
end
```

spec/models/dataset_spec.rb
```diff
@@ -123,4 +123,19 @@
       expect(dataset.public_updated_at).to eq(dataset.updated_at)
     end
   end
+
+  describe "#released" do
+    it "returns true when there is a link" do
+      dataset = FactoryGirl.create(:dataset, datafiles: [FactoryGirl.create(:datafile)])
```
andrewgarner: How about using the `create_list` or `create_pair` method from FactoryGirl to make this a bit more expressive?

```rb
dataset = FactoryGirl.create(:dataset, datafiles: FactoryGirl.create_list(:datafile, 1))
```

See https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#building-or-creating-multiple-records

We could also maybe move this into the Factory definition using associations so we could have a `dataset_with_datafiles` factory – https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#associations

benthorner: The list technique is better, but I'm against having factories that aren't directly related to a single model - it reduces visibility too much for me.

app/models/dataset.rb
benthorner: But it makes the file so long :-( I notice we don't take this approach anywhere with attr_xxx, so it's somewhat consistent with that.

spec/models/dataset_spec.rb
andrewgarner: It makes it easier to focus on the detail of the test, and not the inconsequential detail of object creation.

spec/models/dataset_spec.rb
benthorner: I like to avoid using contexts unless they actually add context; in this case, we're just changing the subject. It also results in more indentation meandering as you try to read the file. I used to do contexts a lot, but then found I was typing a lot of boilerplate for no benefit.

With the predicate matcher, I'm tempted but I'm thinking it looks a bit magic for someone coming to read the tests - unless you're very familiar with RSpec, using a smaller vocabulary is more friendly to the reader.

app/models/dataset.rb
andrewgarner: That's why you use Vim. I agree that the convention isn't universally adopted, but I think it's a good one to adopt as we go along.

spec/models/dataset_spec.rb
benthorner: I think I'll compromise and do what I did with find - use a trait.

app/models/dataset.rb
benthorner: Hmm...not sure how a particular editor helps with this. In my previous job, we usually managed to keep each file below a page in length. For me, it's helpful to be able to clearly see that a file in one go as much as possible.

benthorner: As this is mainly stylistic, perhaps we should get a third opinion - I'm happy to go with a majority.

rossjones: I can't read that crammed onto four lines :(

benthorner: I guess the alternative should be to move this code into it's own presenter, which is it's proper place. But not now.

benthorner:
Just so it's clear, https://github.com/alphagov/datagovuk_find/pull/613 is a prerequisite to this PR.

app/models/dataset.rb
```diff
@@ -61,8 +61,7 @@ def publish!
   # What we actually want to index in Elastic, rather than the whole
   # dataset.
   def as_indexed_json(_options = {})
-    as_json(
-      only: %i[
+    as_json(only: %i[
```
andrewgarner: What are we gaining by changing this?

benthorner: I don't recall any style guide ever suggesting having a hanging method bracket being a good practice.

benthorner: I think it's fair to make a small improvement on the same piece of code, especially as it doesn't modify the diff. 

Perhaps we should do some pairing next week so that we understand each others' expectations better.

PR #635, opened 2018-05-18
Fixes some problems with non-perfectly-formed URLs

rossjones:
RestClient doesn't like making requests with non-encoded strings, and so
this PR will encode the string, if it is not already. This stops the
validation complaining about URLs the user has given that may contain
spaces.

Also adds ftp to the list of allowed URLs.

### DIFF
--- a/lib/validators/url_validator.rb
+++ b/lib/validators/url_validator.rb
@@ -31,12 +31,12 @@ class UrlValidator < ActiveModel::Validator
   end
 
   def encoded?(original_url)
-    decoded = Addressable::URI.unescape(original_url)
+    decoded = Addressable::URI.unencode(original_url)
     decoded != original_url
   end
 
   def encoded_url(url)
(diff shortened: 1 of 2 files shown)


lib/validators/url_validator.rb
```diff
@@ -17,18 +18,31 @@ def url_present?(record)
...
+
+    if %w(http https ftp).exclude? uri.scheme
       create_validation_error(record, error)
     else
       true
     end
   end
 
+  def encoded?(original_url)
+    decoded = Addressable::URI.unescape(original_url)
```
andrewgarner: `Addressable::URI` has `unencode` which is aliased to `unescape`. Perhaps we should use "encode" consistently?

rossjones: Yup, can do. I wonder why it doesn't have `decode` though

rossjones: @andrewgarner Done

PR #639, opened 2018-05-30
Remove the UrlValidator from the Link model

rossjones:
The UrlValidator is problematic for several reasons:

 * It occassionally complains about invalid URLs that are perfectly valid
 * It uses a HEAD request which may not be supported by the remote server
 * It only supports checking HTTP urls (although FTP ones are valid)
 * Links are not checked for malicious content

These issues cause the current sync process to break and stop processing
datasets, meaning that we end up with lots of non-synced datasets in
legacy that need to be manually imported (after disabling the
validator).

As the longer term plan is to switch to the GOV.UK link checker, this
feature has been removed until that work can be undertaken.


benthorner:
There's also a line in config/application that can go :-)

andrewgarner:
Could we keep the validator, but only apply it in the context of manually creating a dataset? There is prior art for this – https://github.com/alphagov/datagovuk_publish/commit/55d293a2534775fe02ed677d483c0634f0765d55

PR #642, opened 2018-06-01
Add stub worker to coordinate full CKAN sync

benthorner:
https://trello.com/c/vmcuCKQ8/263-work-out-which-datasets-to-sync-using-the-old-ckan-api-26

The aim of this work is to create a list of which datasets need to be
created, updated or deleted by doing a full comparison against the CKAN
API. Everything is namespaced to allow for multiple CKAN API versions

In order to keep the acceptance tests realistic, the worker also
does a minimal amount to get them passing, pending a fuller
implementation that sets all of the appropriate attributes correctly.

Deletion is determined by the absence of a package in the results
returned from CKAN. In order for this to work with manually published
datasets - and noting a user cannot delete a dataset in Publish - the
`legacy_name` field is used to isolate datasets originating from CKAN.

The CKAN client is made up of modules in order to keep the files
readable, but also to make it easier to add the same capabilities to
future clients (for other versions of the CKAN API). The client does not
implement any retry mechanism, as this should be handled naturally by
the job framework that will be executing this code. The client does not
customize any timeouts, as manual testing indicates this is unnecessary.


app/services/ckan/v26/version_diff.rb
```diff
@@ -0,0 +1,50 @@
...
+          create: diff_create(packages, datasets),
+          update: diff_update(packages, datasets),
+          delete: diff_delete(packages, datasets)
+        }
+      end
+
+    private
+
+      def diff_create(packages, datasets)
+        dataset_uuids = Set[*Dataset.pluck(:uuid)]
```
andrewgarner: Did you intent to use the `datasets` scope here?

app/workers/ckan_sync_worker.rb
```diff
@@ -0,0 +1,31 @@
...
+    create_new_datasets(actions[:create])
+    update_existing_datasets(actions[:update])
+    delete_old_datasets(actions[:delete])
+  end
+
+private
+
+  def create_new_datasets(packages)
+    packages.each do |package|
+      Dataset.create(title: "title",
```
andrewgarner: I wonder whether we should extract this into a separate job? If any of the datasets fail to create, update or delete then whole job will fail. If we break the work down into individual jobs the failure will only effect the isolated dataset and not block the whole batch.

PR #650, opened 2018-06-07
Implement new import worker

benthorner:
https://trello.com/c/73zsOucF/262-sync-a-single-dataset-using-the-ckan-api-26

The aim of this PR is to replace the stubs used by the CKAN sync worker to create and update a dataset (and it's relations). Each create or update action is now done by a new CKAN import worker, which is creates a dataset if it does not exist and any case overwrites all of its attributes. Any relations (inspire dataset, links) are created, updated or deleted to match the state of the package.

The way some fields are populated have changed by comparison to the existing sync. Some fields are not populated (e.g. secondary topics) as the values are not used, while some have moved elsewhere (e.g. last_updated_at -> updated_at). We should verify the integrity of the sync by doing a random sample, once the elasticsearch integration is working.

Running the worker successfully imports all ~45,000 datasets in about 30 minutes, with subsequent updates taking around 2 minutes to complete. There are 27 datasets that repeatedly do not import successfully, which appears to be because they do not properly exist in CKAN. I will investigate and ensure these are cleaned up, although they won't interfere with the job running in production.

### DIFF
--- a/app/services/ckan/v26/dataset_importer.rb
+++ b/app/services/ckan/v26/dataset_importer.rb
@@ -2,17 +2,9 @@ module CKAN
   module V26
     class DatasetImporter
       def call(dataset, package)
-        dataset.title = "title"
-        dataset.summary = "summary"
-        dataset.last_updated_at = package["metadata_modified"]
-        dataset.organisation = find_organisation(package)
+        attributes = DatasetMapper.new.call(package)
+        dataset.assign_attributes(attributes)
         dataset.save
-      end
-
-    private
-
-      def find_organisation(package)
-        Organisation.find_by(uuid: package["owner_org"])
       end
     end
   end
--- a/app/services/ckan/v26/dataset_importer.rb
+++ b/app/services/ckan/v26/dataset_importer.rb
@@ -1,11 +1,12 @@
 module CKAN
   module V26
-    class DatasetMapper
+    class DatasetImporter
       def call(dataset, package)
         dataset.title = "title"
         dataset.summary = "summary"
         dataset.last_updated_at = package["metadata_modified"]
         dataset.organisation = find_organisation(package)
+        dataset.save
       end
 
     private
--- a/app/services/ckan/v26/inspire_updater.rb
+++ b/app/services/ckan/v26/inspire_updater.rb
@@ -1,6 +1,6 @@
 module CKAN
   module V26
-    class InspireImporter
+    class InspireUpdater
       def call(dataset, package)
         if inspire?(package)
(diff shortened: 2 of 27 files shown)


app/services/ckan/v26/dataset_importer.rb
```diff
@@ -0,0 +1,11 @@
+module CKAN
+  module V26
+    class DatasetImporter
```
andrewgarner: Maybe this should be `PackageImporter`? We're "importing" the package and "mapping" it to a dataset?

benthorner: I had the same thought. My conclusion was that the files are organised by the local data structures (datasets, inspire_datasets and links); for inspire_datasets there's no equivalent model terminology in CKAN. Perhaps 'DatasetUpdater' would be less ambiguous?

andrewgarner: +1 for `DatasetUpdater`

app/services/ckan/v26/inspire_updater.rb
```diff
@@ -0,0 +1,31 @@
+module CKAN
+  module V26
+    class InspireUpdater
+      def call(dataset, package)
```
andrewgarner: Perhaps we could provide just the id of the dataset instead of the whole object?

benthorner: I was going more for uniformity and avoiding the import worker knowing too much about what the individual updaters need. It's possible the inspire updater would need to see the dataset attributes.

app/services/ckan/v26/version_diff.rb
```diff
@@ -20,25 +22,28 @@ def call
...
       def diff_delete(packages, datasets)
-        package_uuids = packages.map { |package| package["id"] }
+        package_uuids = packages.map { |package| package.get("id") }
         datasets.where.not(uuid: package_uuids)
       end
 
-      def package_is_changed?(package, last_updated_at)
-        last_updated_at.iso8601 < package["metadata_modified"]
+      def package_is_changed?(package, updated_at)
+        updated_at.iso8601 < package.get("metadata_modified")
```
andrewgarner: Should this be `DateTime.parse(package.get("metadata_modified"))`? Our package is just a hash?

benthorner: It turns out `.parse` is quite expensive, whereas part of the beautify of ISO8601 is that the timestamps are lexicographically comparable, as here. I put it in a method to keep the optimisation separate.

PR #653, opened 2018-06-07
Enable deletion of datasets in ES (with tests)

benthorner:
https://trello.com/c/73zsOucF/262-sync-a-single-dataset-using-the-ckan-api-26

Deletion is now supported as part of the renamed ESPublishWorker (the
new name helping to differentiate with other import/export workers).

This highlighted a lack of tests for ES publishing (create, update and
delete). In order to make this easier, the dataset as_indexed_json
method now exports all dataset attributes (noting Find is now able to
cope with this). Pulling the testing out to the feature level then made
some of the lower-level specs redundant, so these have been removed.

A spec helper has also been added to cleanup the test index, which was
previously growing without bound.


spec/spec_helper.rb
```diff
@@ -19,6 +19,10 @@
 
   config.order = :random
 
+  config.before(:each) do
```
andrewgarner: There is already a `config.before(:each)` definition?

PR #668, opened 2018-06-13
Fix publish failure due to large summaries

benthorner:
https://trello.com/c/wpH2Jdt3/356-sync-a-single-dataset-to-find-using-the-ckan-api-26

A couple of documents have large summaries that are preventing them from
being indexed. We should disable this behaviour for large summaries as
it's unlikely the indexing will be beneficial for so much text.

https://www.elastic.co/guide/en/elasticsearch/reference/current/ignore-above.html


benthorner:
Looks like you were right ;-)

PR #676, opened 2018-06-18
Remove redundant delete task (now via UI)

benthorner:
https://trello.com/c/EOv1EI5a/326-cleanup-sync-rake-tasks


benthorner:
If it helps, running this task is now futile as the sync will just replace whatever it tried to manually delete. Also, we already merged the PR to delete the DatasetIndexService this thing uses :-|

PR #679, opened 2018-06-18
Remove legacy code for CKAN package sync

benthorner:
https://trello.com/c/g07wP2X9/325-remove-old-sync-code-after-replacing-it-with-the-new-sync-code

The previous sync mechanism is no longer used and can be safely deleted.
There is still some code that is used for importing legacy
organisations, which is why the legacy namespace still remains.

The import script no longer tries to import packages, as this tends to
take a long time and is usually unnecessary. It is still possible to
manually import packages by running the PackageSyncWorker.


bin/import
```diff
@@ -13,20 +13,14 @@ print_error() {
...
 rails import:locations["lib/seeds/locations.csv"]
 
 print "Adding organisations"
 rails import:legacy_organisations["data.gov.uk-ckan-meta-data-latest.organizations.jsonl"]
 
-print "Adding datasets"
-rails import:legacy_datasets["data.gov.uk-ckan-meta-data-latest.v2.jsonl"]
-
 print "Cleaning up"
 rm data.gov.uk-ckan-meta-data-latest.organizations.jsonl data.gov.uk-ckan-meta-data-latest.v2.jsonl
```
andrewgarner: `data.gov.uk-ckan-meta-data-latest.v2.jsonl` can be removed from this cleanup step too.

benthorner: Done - thanks for spotting :-)

PR #688, opened 2018-06-19
Simplify the elastic config (less code, more YML)

benthorner:
https://trello.com/c/RouPcOX9/400-simplify-elastic-configuration-logic

The ES initialization logic was confusing because it was coupled to the
associated YML file. Making the YML file do more work helps to clarify
where the configuration is coming from in each environment.

It should not be necessary to verify the ES certificate in the same way
that we don't verify the identity of the Postgres DB that the app is
connecting to.


benthorner:
Closing as this is only desirable.

PR #693, opened 2018-06-20
Catch ES errors and retry when reindexing

benthorner:


app/services/datasets_indexer_service.rb
```diff
@@ -185,6 +185,9 @@ def bulk_index(datasets)
       type: ::Dataset.__elasticsearch__.document_type,
       body: prepare_records(datasets)
     )
+  rescue Elasticsearch::Transport::Transport::Error => e
+    logger.warn(e)
+    bulk_index(datasets)
```
andrewgarner: I think we can use `retry` here?

benthorner: Done :-)

PR #699, opened 2018-06-21
Remove scripts directory

benthorner:
https://trello.com/c/3r6gUCTG/405-rewrite-the-dgu-developer-docs

This directory contained scripts for deployment, replication and setting
CF environment variables. The developer docs now include the commands to
perform a manual deployment; we should not be modifying CF environment
variables as a common task - they should be controlled using Terraform.
Replication from a staging/production DB has not been a common task.


scripts/replication.sh
```diff
@@ -1,13 +0,0 @@
-#!/usr/bin/env bash
```
andrewgarner: Are these steps documented anywhere?

This strategy is very handy to bootstrap the application without having to manually run a full import against the CKAN API.

scripts/deploy.sh
```diff
@@ -1,65 +0,0 @@
-#!/bin/bash
```
andrewgarner: This is required for deployments – https://github.com/alphagov/datagovuk_publish/blob/6c7654d0d0f646ba5bc5f859cfe58978b25b82c6/.travis.yml#L33

PR #706, opened 2018-06-25
Remove unused Rack::Cors plugin

benthorner:
https://trello.com/c/DUMGb0Td/427-remove-unused-publish-features

This is unused as Publish is no longer accessed as robotically.


config/application.rb
```diff
@@ -40,22 +40,10 @@ class Application < Rails::Application
...
-      end
-    end
-
     config.autoload_paths += [Rails.root.join("app/workers")]
     config.autoload_paths += [Rails.root.join("lib/validators")]
     config.autoload_paths += [Rails.root.join("lib/ckan")]
 
     config.elasticsearch = config_for(:elasticsearch)
-
-    config.filter_parameters << :password
```
andrewgarner: These are duplicate in https://github.com/alphagov/datagovuk_publish/blob/master/config/initializers/filter_parameter_logging.rb too.

Are you removing them because we no longer have a form with `password` fields?

benthorner: We should remove this one as it's custom. The initializer comes from `rails g`, so I'd rather leave it.

andrewgarner: 👍

PR #712, opened 2018-07-11
Avoid running db:create as part of the manifest

benthorner:


benthorner:
Nope - this series of PRs is an attempt remove a mess of migrations someone did in the past. Hopefully after this it'll become stable.

PR #716, opened 2018-08-14
Add indexes to two columns in the links table

brucebolt:


db/migrate/20180814160005_add_link_indexes.rb
```diff
@@ -0,0 +1,6 @@
+class AddLinkIndexes < ActiveRecord::Migration[5.1]
+  def change
+    add_index :links, :type
+    add_index :links, :dataset_id
```
kevindew: You'll probably want to add these indexes [concurrently](https://github.com/ankane/strong_migrations#dangerous-operations)

Which can be done by adjusting your migration like this: https://github.com/alphagov/publishing-api/blob/e6cef33bae7959f55a3876dbe24353dad6793e68/db/migrate/20180801195938_add_indexes_for_linkables.rb#L2-L6

PR #718, opened 2018-08-15
Limit concurrency and retries of sidekiq workers

elliotcm:
The issue:
The sync queue for DGU has been found to clog up, with sometimes 
millions of import jobs queued.  Expectation of number of import jobs 
queued is in the double digits, with 46k being the absolute maximum as 
that's the number of datasets.

The cause:
It has been observed that during these blockages, most or all 5 
processes are working on sync jobs, with import jobs trickling through 
or not at all.  With new sync jobs appearing every 10 minutes and import 
jobs taking 3 weeks to expire, this causes a mildly exponential increase 
in queued jobs.  Current hypothesis is that the multiple workers 
requesting 46k dataset slugs in pages of 1000 each is slowing down CKAN 
and causing the backup.  Since only sync jobs are being processed, more 
import workers are queued continuously.

This solution:
By limiting sync jobs to 1 concurrent worker we should allow import jobs 
to be processed regardless of the number, with sync jobs waiting 
patiently for their turn.  Expiring the (transient) import workers after 
a few tries means they won't hang around for weeks if there is an issue.

https://trello.com/c/oxh2ekJr/382-spike-2-days-find-out-how-we-might-stop-the-datagovuk-sync-queue-from-backing-up-mandatory-%F0%9F%8D%90


app/workers/ckan/v26/ckan_org_sync_worker.rb
```diff
@@ -2,6 +2,7 @@ module CKAN
   module V26
     class CKANOrgSyncWorker
       include Sidekiq::Worker
+      sidekiq_options queue: :sync
```
barrucadu: This worker is scheduled daily, so does it also need a retry limit perhaps?

The table here https://github.com/mperham/sidekiq/wiki/Error-Handling suggests that 13 or 14 might be a good limit.

brucebolt: This one syncs organisation metadata across from CKAN.  I wonder whether it should actually be run more often than daily?  Would a new publisher have to wait up to a day for their datasets to show on Find?

elliotcm: 14 would be 10 hours, right?

elliotcm: @brucebolt I don't have a good answer to that, requires more domain knowledge than I have, but I don't at first glance see why not.

barrucadu: The table under "This table contains approximate retry waiting times (click to expand)." gives a total wait time for 14 as 1d  1h 17m 36s.

elliotcm: Ah, I was looking at the formula for the next backoff duration.  I've set it to 13.

elliotcm:
Do not merge yet, we're going to deploy this branch to staging for a while.

elliotcm:
Running on staging now, seems OK, we'll ship on Monday if no giant problems.

elliotcm:
Tested on staging over the weekend, all seems fine.

PR #719, opened 2018-08-16
Expose Sidekiq queue metrics to Prometheus

brucebolt:
Uses the `sidekiq-prometheus-exporter` gem to make queue metrics available to Prometheus.

They are available at localhost:8080/sidekiq/metrics, this is only available locally or
by creating a proxy through to the app server.


Gemfile
```diff
@@ -26,6 +26,7 @@ gem 'rest-client', '~> 2.0.2'
 gem 'rubyzip', '~> 1.2'
 gem 'sass-rails', '~> 5.0'
 gem 'sentry-raven'
+gem 'sidekiq-prometheus-exporter'
```
thomasleese: Might be nice to specify a `~>` version here.

thomasleese:
@idavidmcdonald We've got this PR up to add a Prometheus `/metrics` endpoint to the DGU application, does this look good to you?

Gemfile
brucebolt: Good point.  I've updated this with a version.

idavidmcdonald:
@thomasleese I'll take a look

idavidmcdonald:
I think this approach might not work or may need some adapting (sorry). 

For some context, the way we scrape apps which run on the PaaS is that an application needs a web page accessible to the public internet available at /metrics (as our Prometheus does not run in the PaaS, only service discovery does and private networking does not exist in the PaaS). Then your application is configured to be scraped as per https://reliability-engineering.cloudapps.digital/manuals/monitor-paas-app-with-prometheus.html#bind-your-apps-to-the-prometheus-service.

On top of this, teams generally have been concerned about making their metrics available to the public internet without some sort of authentication in place. We used to go about this by teams using GDS specific prometheus web app clients that provided some http basic auth infront of your metrics endpoint. However we've recently introduced a PaaS called re-ip-whitelist as per https://reliability-engineering.cloudapps.digital/manuals/monitor-paas-app-with-prometheus.html#bind-your-apps-to-the-prometheus-service. Note, I think the docs could be a bit more explicit about the explicit behaviour of the whitelist service but it would proxy requests and only allow those through from prometheus and office IPs I believe (I would need to double check that).

In your case, assuming you would not be happy having your metrics available, we would need to use the IP whitelist app. A question on our side to figure out is that if you use this, would it protect a /metrics endpoint or would it infact protect all routes (meaning real users from outside the office couldn't get to your site). A quick way of not answering this question is that we decide it doesn't matter by running the sidekiq metrics exporter as it's own standalone PaaS application connected to your Redis. A current hypothesis is that that would still give you the metrics you require but I'm quite unfamiliar with Sidekiq and Redis so that would need to be tried and confirmed.

Given my unfamiliarity with the DGU publisher, my initial question is does this PR enable a /metrics endpoint available for a PaaS app, ie. `publish-data-beta-production-worker.cloudapps.digital/metrics`? (I assume it is the worker app that this would be for but do correct me if I'm wrong).

My main suggestion however would be to try running a very simple Rack app with the sidekiq metric exporter as per https://github.com/Strech/sidekiq-prometheus-exporter#rack-application. This should provide an app with a public accessible /metrics endpoint that will be easy to bind to our Prometheus without any issues with IP whitelisting. Note however, this relies on my assumption being correct that the metrics from your production Sidekiq app would be the same as from a standalone Sidekiq app used for monitoring - this could be entirely incorrect though so please do tell me if so.


I hope this long long bit of information is helpful. I'm around for the rest of the day to try and advise still but I'm on holiday from Friday-Wed so someone else on the RE Observe team should be able to help whilst I'm away (we have a `re-prometheus-support` channel).

idavidmcdonald:
A quick look shows that the RE-ip-whitelist service currently blocks access for all routes, not just the /metrics endpoint - https://github.com/alphagov/re-paas-ip-whitelist-service/blob/master/nginx.conf. 

Maybe this is something we could change but would probably need some discussion within our team of the implications at the moment.

PR #721, opened 2018-08-21
Make dataset deletion more robust

elliotcm:
Previously, records that were marked as draft in the local Postgres database would never get marked as published, even if they are made available again though the CKAN API.

This updates that logic so that datasets which are returned in the CKAN API response but are draft in the local database are always updated, thus marking them as published again.

This is particularly useful for recovering from a repeated class of incident where the importer mistakenly deletes (makes draft) several thousand packages.

This change also causes the depagination process to raise exceptions if a number of Bad Situations arise, primarily from either misinterpreting CKAN responses or from CKAN providing intact responses which misrepresent reality (e.g. far fewer datasets than are really in CKAN).

This should help avoid a recurring incident we've been seeing where this app accidentally deletes a bunch of published datasets.

https://trello.com/c/smOj0SX5/408-investigate-dgu-package-deletion


.rubocop.yml
```diff
@@ -5,9 +5,13 @@ AllCops:
     - config.ru
     - db/schema.rb
     - vendor/**/*
+    - lib/ckan/v26/depaginator.rb
```
thomasleese: Why do we need to do this?

elliotcm: It's a very tough class to make readable (and not currently worth our while to do a really significant refactor) and every change rubocop flagged would make it harder to read.

elliotcm: Primarily around the heredocs.

elliotcm:
This is running on staging until after the deploy freeze.

PR #725, opened 2018-10-12
Add schema ID to imported packages

elliotcm:
https://trello.com/c/TeIA6PUL/60-implement-organogram-viewer

This is so we can distinguish organogram datasets from others in Find.


brucebolt:
We will have to make a change for new CKAN: `schema` is called `schema-vocabulary` in the API.

PR #727, opened 2018-10-25
Update Publish to reflect changes to new CKAN API

brucebolt:
The new CKAN instance has some slight changes to the API schema.  These commits allow the `harvest_object_id`, `harvested` and `schema_id` values to be populated using the new schema.  Backwards compatibility is maintained to ensure the Publish sync process can read from both the old and new APIs.

Trello card: https://trello.com/c/WFsQ1AcV/65-update-publish-to-use-new-harvestobjectid-and-schema-vocabulary

### DIFF
--- a/app/models/ckan/v26/package.rb
+++ b/app/models/ckan/v26/package.rb
@@ -14,6 +14,11 @@ module CKAN
         @extras[key]
       end
 
+      def get_harvest(key)
+        @harvest ||= hashify(@package["harvest"] || [])
+        @harvest[key]
+      end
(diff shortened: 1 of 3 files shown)


brucebolt:
Yes, I'll put one in

brucebolt:
I've added a test for the package import worker with an API response from new CKAN.  Left the old tests in place to ensure we maintain backward compatibility.

Tests for INSPIRE datasets have been deliberately excluded, since there is a separate story to get this working.

PR #731, opened 2018-12-20
[REQUIRES REINDEX] Revamp the way we calculate public_updated_at

benthorner:
https://govuk.zendesk.com/agent/tickets/3539916

Previously we calculated the public_updated_at field for a dataset based
on the first available timestamp in the list of things to try. This
changes the way we calculate this field as part of the publish payload,
so that we try to get the latest timestamp out of the ones available.

### DIFF
--- a/app/models/dataset.rb
+++ b/app/models/dataset.rb
@@ -104,9 +104,11 @@ class Dataset < ApplicationRecord
   end
 
   def public_updated_at
-    inspire_dataset_reference_date ||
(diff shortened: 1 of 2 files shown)


benthorner:
@brucebolt thanks for pointing out the ticket - fixed :-). Could you expand a bit on what you mean by the INSPIRE v non-INSPIRE date? Just want to make sure I've done the right thing...

PR #766, opened 2019-06-10
Fix organogram filenames

kentsanggds:
## What

It seems like changes to the ckan database and solr index are not picked up by Publish automatically, so in order to force it to pick up the filename fixes in this card - https://trello.com/c/5KrmZ1tb/1025-fix-bug-to-show-visualisation-on-junior-post-organograms, we need to run a one-off script to update the filenames in the Publish DB, then we just reindex elasticsearch and the filenames should be updated.

The input to the process was generated by a run of this script on the appropriate environment - 

https://github.com/alphagov/ckanext-datagovuk/blob/master/bin/python_scripts/fix_organograms_s3_filenames.py

### DIFF
--- a/lib/tasks/fix_existing_organogram_filenames.rake
+++ b/lib/tasks/fix_existing_organogram_filenames.rake
@@ -1,0 +1,6 @@
+require_relative 'update_organogram_filenames.rb'
+
+desc "Fix existing organogram filenames"
+task :fix_existing_organogram_filenames => :environment do
+  UpdateOrganogramFilenames.new().call
+end
--- a/lib/tasks/fix_existing_organogram_filenames.rake
+++ b/lib/tasks/fix_existing_organogram_filenames.rake
@@ -1,6 +1,6 @@
 require_relative 'update_organogram_filenames.rb'
 
 desc "Fix existing organogram filenames"
-task :fix_existing_organogram_filenames => :environment do
-  UpdateOrganogramFilenames.new().call
+task "fix_existing_organogram_filenames" => :environment do
+  UpdateOrganogramFilenames.new.call
 end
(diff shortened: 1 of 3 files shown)


lib/tasks/fix_existing_organogram_filenames.rake
```diff
@@ -0,0 +1,6 @@
+require_relative 'update_organogram_filenames.rb'
+
+desc "Fix existing organogram filenames"
+task :fix_existing_organogram_filenames => :environment do
+  UpdateOrganogramFilenames.new().call
```
thomasleese: Generally in Ruby if a function doesn't take any arguments you would call it without the `()`.

```suggestion
  UpdateOrganogramFilenames.new.call
```

lib/tasks/update_organogram_filenames.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    parse_csv(csv_file)
+    msg = replace_urls()
+    puts msg
+
+  end
+
+  def parse_csv(csv_file)
+    puts "Parsing CSV '#{csv_file}'"
+    File.open(csv_file).readlines.each do |line|
```
thomasleese: You can read this using the CSV module.

```rb
CSV.foreach(csv_file) do |row|
  @old_urls << row.first
  @new_urls << row.last
end
```

lib/tasks/update_organogram_filenames.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+  end
+
+  def replace_urls
+    puts "Searching for urls containing '-posts-'..."
+
+    if @old_urls.empty?
+      "No urls to process"
+    else
+      Link.all.each do |link|
```
thomasleese: If you're interested, using [`Link.find_each` is slightly more efficient here](https://api.rubyonrails.org/classes/ActiveRecord/Batches.html#method-i-find_each), but it probably don't make a big difference here.

lib/tasks/update_organogram_filenames.rb
kentsanggds: I'll try that

PR #803, opened 2019-07-01
Update README with some PaaS info

kentsanggds:
## What

Commands will not run without starting the shell up in the correct build environment, so it needs to be run each time if we want to do things like re-index elasticsearch.


thomasleese:
@kentsanggds I've rebased this so the tests aren't failing with linting errors now.

PR #893, opened 2020-01-23
Remove style variables

deborahchua:
This app use the `govuk_frontend_toolkit` for its styles,
but recently has been having trouble finding variables such as
$link-colour, which is blocking deployments.

As we don't need to access Publish via the frontend we can
remove references to these variables in the affected Sass files.

Trello card: https://trello.com/c/LhIWTWKQ/1693-3-remove-dependency-which-is-blocking-deployment-of-dgu-publish

### DIFF
--- a/app/assets/stylesheets/application.css.scss
+++ b/app/assets/stylesheets/application.css.scss
@@ -53,4 +53,3 @@ $alpine-fresh: #128400;
 @import 'elements/components';
 
 @import 'govuk-overrides';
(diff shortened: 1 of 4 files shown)


deborahchua:
I did consider this but was concerned it would break other things, but @kentsanggds can decide whether we should prioritise removing all frontend stuff from this app.  Hopefully this will fix the current issue! 🤞

PR #921, opened 2020-03-02
Remove `lint` Rake task

issyl0:
- Now we don't use `govuk-lint` to wrap RuboCop, we _could_ switch this
  to run `bundle exec rubocop`, but that's not very useful as it hides
  the directories that it operates on, and we're trying to move away
  from wrappers.
- Instead, I think we can leave people to run `bundle exec rubocop` on
  their own - it's less typing than `bundle exec rake lint` and
  `rubocop --help` works for options.
- Now there's no `lint` Rake task, we have to run `rubocop` on CI
  directly.

https://trello.com/c/IEUuFOVa/129-remove-the-lint-rake-task-from-apps-that-have-it as part of the developer tooling group

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_script:
 script:
   - bundle exec rails db:create
   - bundle exec rails db:schema:load
+  - bundle exec rubocop app spec lib Gemfile
   - bundle exec rake
   - bundle exec codeclimate-test-reporter
 env:
--- a/Rakefile
+++ b/Rakefile
@@ -5,4 +5,4 @@ require_relative 'config/application'
 
 Rails.application.load_tasks
 
-task default: %i(spec lint brakeman)
+task default: %i(spec brakeman)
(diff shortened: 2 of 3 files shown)


.travis.yml
```diff
@@ -16,6 +16,7 @@ before_script:
 script:
   - bundle exec rails db:create
   - bundle exec rails db:schema:load
+  - bundle exec rubocop app spec lib Gemfile
```
thomasleese:
```suggestion
  - bundle exec rubocop app spec lib Gemfile config
```

Or can we run `bundle exec rubocop` and get the default directories? Or was this done to match the existing behaviour and then we can fix it later (I'm guessing there might be linting issues in the `config` directory?).

issyl0: When I did this, I didn't have the idea to do the `rubocop-govuk` changes yet (it was only after several goes at this that it got tedious enough). And this was done to match Jenkins (even though it's not Jenkins, but consistency).

issyl0: This will get a RuboCop update in due course and in doing that we can remove the directories from this line.

PR #979, opened 2020-05-20
Add Dockerfile, setup-docker and docker environment

kentsanggds:
## What

In order to get Publish to run in the alphgov/docker-ckan
dev stack it is necessary to add these files and add in environment vars for
redis and ckan hosts so that it is still possible to run Publish as a standalone app.

### DIFF
--- a/bin/setup-docker.sh
+++ b/bin/setup-docker.sh
@@ -1,0 +1,26 @@
+#!/bin/bash
+# This setup script is used with the docker-ckan dev stack
+
+echo '===== Setup datagovuk publish ====='
+
+RETRIES=10
+until psql -h db -U ckan -d ckan -c "select 1" > /dev/null 2>&1 || [ $RETRIES -eq 0 ]; do
+    echo "Waiting for postgres server, $((RETRIES--)) remaining attempts..."
+    sleep 1
+done
+
+bin/rails db:setup
+rails db:environment:set RAILS_ENV=development
+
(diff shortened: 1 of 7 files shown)


bin/setup-docker.sh
```diff
@@ -0,0 +1,26 @@
...
+echo '===== Setup datagovuk publish ====='
+
+RETRIES=10
+until psql -h db -U ckan -d ckan -c "select 1" > /dev/null 2>&1 || [ $RETRIES -eq 0 ]; do
+    echo "Waiting for postgres server, $((RETRIES--)) remaining attempts..."
+    sleep 1
+done
+
+bin/rails db:setup
+rails db:environment:set RAILS_ENV=docker
```
thomasleese, line 13: I think we should avoid adding a new environment for Docker like this as it means we need to maintain an extra file in `config/environments` and it would also be quite surprising for new developers I think as usually there's only a `development`, `production` and `test` environment.

For the things we need to customise like the elasticsearch host and `ckan_v26_base_url` could we use environment variables?

kentsanggds: hmmm, ok, I'll see if I can set up env vars instead

config/environments/docker.rb
```diff
@@ -0,0 +1,5 @@
+require_relative "./development"
```
thomasleese, line 1: I think we can remove this file now!

PR #1009, opened 2020-08-18
Temporarily increase max deletions to 500

kentsanggds:
## What

In order to run the CKAN functional tests, some existing test data had to be removed which may have caused the MAX_DELETIONS threshold to be hit, so temporarily increase it to 500 to allow test data to be removed.


risicle (approved):
As long as we have a ticket to put it back

PR #1027, opened 2020-10-14
Move code to app directory and remove autoload config

pixeltrix:
Autoloading code in production is generally unsafe and when mixing the namespace across autoloaded paths, eager loaded paths and manual require commands it can result in unpredictable behaviour. By moving all the code into the app directory we can be certain that the code is eager loaded before Sidekiq processes any jobs.


pixeltrix:
Creating this as a draft to see if it'll pass on Travis because it doesn't run with govuk-docker

thomasleese (approved):

PR #1035, opened 2020-11-16
Upgrade elasticsearch to 7.9.0

kentsanggds:
## What

Upgrade elasticsearch to 7.9 as elasticsearch 6 will be no longer supported on PaaS.

## Reference 

https://trello.com/c/xjutCWHA/2215-upgrade-elasticsearch-version-on-paas


staging-worker-manifest.yml
```diff
@@ -1,5 +1,5 @@
 applications:
-- name: publish-data-beta-staging-worker
+- name: publish-beta-staging-worker
```
thomasleese, line 2: Finally! 😆

thomasleese (approved):
I don't fully understand the differences in this version of Elasticsearch, but it looks sensible and the tests pass.

PR #1037, opened 2020-11-24
Add ci.yml for github actions

kentsanggds:
## What

Added github action test to allow for a faster test of PRs  as travis is no longer working.

## Reference 

https://trello.com/c/3MHhGs5C/2602-replace-travis-ci-with-github-actions-on-publish-and-find

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,54 @@
+on: [pull_request]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    services:
+      postgres:
+        image: postgres:9.5
+        env:
+          POSTGRES_HOST_AUTH_METHOD: trust
+        ports:
+          - 5432:5432
+        options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
+      redis:
+        image: redis
+        ports:
+          - 6379:6379
+        options: --health-cmd "redis-cli ping" --health-interval 10s --health-timeout 5s --health-retries 5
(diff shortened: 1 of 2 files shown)


config/database.yml
```diff
@@ -1,25 +0,0 @@
-default: &default
```
1pretz1, line 1: Has this config been deleted intentionally?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
+on: [pull_request]
```
1pretz1, line 1: Should we change this to "push" rather than "pull request"? I find it useful being able to have the tests run without opening a PR. It would also make this consistent with other GOV.UK apps.

1pretz1 (commented):
Looks good Ken, just left a couple of comments.

config/database.yml
kentsanggds: No it probably shouldn't be deleted, I'll add it back in

.github/workflows/ci.yml
kentsanggds: I've based the ci.yml on this PR in Find which was approved - https://github.com/alphagov/datagovuk_find/pull/859

Intention is to only run the tests following the creating of a PR, additional pushes to the PR will also trigger builds. 

Later on release tags will also trigger builds for deploys.

1pretz1: Okay dokay, cheers for the explanation!

1pretz1 (approved):

PR #1076, opened 2021-07-01
Add ci deploy

kentsanggds:
## What

Add deployment steps to `ci.yml` and also update the README to make the steps clearer for other govuk devs.

## Reference 

https://trello.com/c/3MHhGs5C/2602-replace-travis-ci-with-github-actions-on-publish-and-find


.github/workflows/ci.yml
```diff
@@ -52,3 +58,57 @@ jobs:
       - run: bundle exec rails db:schema:load
       - name: Run tests
         run: bundle exec rake
+  deploy_to_staging:
```
ChrisBAshton, line 61: I was going to suggest DRY'ing this up by creating a `install_deployment_dependencies` step which does the cloud foundry / Bundle steps whether we're deploying to Staging or Production. But then I remembered we can't re-use dependencies between steps.

I also wondered if there was a native way to alias different shared steps, but it looks like [GitHub is still working on that](https://github.com/actions/runner/issues/646). So this looks about as good as it's going to get!

README.md
```diff
@@ -5,6 +5,16 @@
 
 This repository contains the beta-stage publishing component of data.gov.uk.
 
+# Deployment
+
+Continuous Integration has been setup using github actions.
```
ChrisBAshton, line 10:
```suggestion
Continuous Integration has been setup using GitHub Actions.
```

README.md
```diff
@@ -5,6 +5,16 @@
 
 This repository contains the beta-stage publishing component of data.gov.uk.
 
+# Deployment
+
+Continuous Integration has been setup using github actions. 
+Tests are run on pull requests, deployments to Staging are via merges of branches into the `main` branch.
```
ChrisBAshton, line 11:
```suggestion
Tests are run on pull requests. Deployments to Staging happen automatically when merging branches into `main`.
```

README.md
```diff
@@ -5,6 +5,16 @@
...
+In order to carry out a release to production a developer in the govuk team will need to create a release tag that leads with a `v` and approve of the deployment in github actions.
```
ChrisBAshton, line 12: I didn't understand "approve of the deployment in github actions" - I don't think GitHub Actions are involved in this step. I've rewritten what I think you meant below:

```suggestion
To deploy to Production, you'll need to [publish a Release](https://github.com/alphagov/datagovuk_publish/releases/new), specifying a "Tag version" beginning with `v` (e.g. `v1.0.0`).
```

README.md
```diff
@@ -5,6 +5,16 @@
...
+
+Further information about the deploying to PaaS are in the developer documents here: 
+
+https://docs.publishing.service.gov.uk/manual/data-gov-uk-deployment.html#paas-staging-and-production-environments
```
ChrisBAshton, line 16: Out of scope, but could the contents of that document be brought into the `docs/` folder of this repo? That way we can keep all the documentation here in one place (helping it stay up to date), and it will also be auto-imported into the Developer Docs.

ChrisBAshton (commented):
LGTM 👍 mostly non-blocking comments below, apart from one confusing thing about the Production deploy that I think needs fixing.

README.md
kentsanggds: There is an additional step to approve a deployment, this is to protect secrets from leaking out as the `production` secrets cannot be protected with tagged releases.

README.md
kentsanggds: I think the reason it's there is so that there's only one place where that information needs to be updated.

ChrisBAshton (approved):
LGTM 👍

PR #1081, opened 2021-08-05
Update package test as test data has changed

kentsanggds:
## What 

Fix the package test as it wasn't updated when the test data structure was updated.

Also fix linting issues due to updated rubocop

## Reference 

https://trello.com/c/DZJFi8z5/2613-create-prs-to-prep-for-29-upgrade


hannako (approved):
👍

PR #1104, opened 2022-02-01
Upgrade Ruby to 2.7.5

ChrisBAshton:
Upgrades Ruby to 2.7.5. See commits for more details.
            This PR should be tested before merge, as per https://docs.publishing.service.gov.uk/manual/ruby.html#update-ruby-version-in-the-relevant-repos.


bilbof (approved):
LGTM

app/lib/ckan/v26/depaginator.rb
```diff
@@ -18,60 +17,63 @@ def initialize(base_url, existing_total:)
       def depaginate
```
sengi, line 17: General comment: there's a bit much nesting in this function; I found it quite hard to read (especially the diff). It was already like that though so I get it if you don't fancy refactoring it right now.

sengi (approved):
Optional: consider adding tests to cover the error handling behaviour.

PR #1208, opened 2023-11-10
Add build image files

kentsanggds:
Build the datagovuk_publish image files for use by the datatgovuk kubernetes cluster.

## Reference

https://trello.com/c/nz3X804Y/1237-deploy-dgu-paas-apps-onto-integration-cluster

### DIFF
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -27,7 +27,7 @@ jobs:
           registry: ghcr.io
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           ref: ${{ inputs.gitRef }}
       - name: Build images (without pushing to registry)
--- a/.github/workflows/build-image.yaml
+++ b/.github/workflows/build-image.yaml
@@ -1,0 +1,45 @@
+name: Build and push images
+
+on:
+  workflow_dispatch:
+    inputs:
+      buildType:
+        description: Decide on build type
+        required: true
+        type: choice
+        options:
+          - build_push
+          - build_only
+  push:
+    branches:
+      - main
+
+jobs:
+  build_and_push:
+    name: datagovuk_publish
+    runs-on: ubuntu-latest
+    permissions:
+      packages: write
+    steps:
+      - name: Login to GHCR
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM ruby:2.7.5
+FROM ruby:3.2.0
```
sengi, line 1: This is an obsolete patch version — 3.2.2. is current.

docker/build-image.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+set -eux
+
+build () {
+  if [ "${ARCH}" = "amd64" ]; then
+    docker build . -t "ghcr.io/alphagov/${APP}:${1}" -f "docker/Dockerfile"
+  else
+    docker buildx build --platform "linux/${ARCH}" . -t "ghcr.io/alphagov/${APP}:${1}" -f "docker/Dockerfile"
```
sengi, line 9: Why not use this for all archs and avoid special-casing amd64 above?

.github/workflows/build-image.yaml
```diff
@@ -0,0 +1,45 @@
...
+    permissions:
+      packages: write
+    steps:
+      - name: Login to GHCR
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a  # v2.1.0
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/checkout@v3
```
sengi, line 30: checkout v4 is current.

docker/build-image.sh
sengi, line 1: optional: prob not worth doing right now but I'd consider using [actions/build-and-push-docker-images](https://github.com/marketplace/actions/build-and-push-docker-images) rather than maintaining another shell script.

sengi (approved):

docker/Dockerfile
kentsanggds: ok, will update

.github/workflows/build-image.yaml
kentsanggds: ok, I'll try the update

docker/build-image.sh
kentsanggds: Hmmmm, I'll have a quick look into that

kentsanggds: I'll stick with what we have for now, that looks like it is pushing images up to dockerhub

PR #1215, opened 2023-11-22
Add build-image-on-tags GitHub action

nimalank7:
Description:
- Adds a GitHub Action to build the docker image but tagged
- Use 'k' instead of 'v' to avoid clashing with the existing docker images as this isn't ready to release without testing yet


kentsanggds (approved):
LGTM

PR #1217, opened 2023-11-22
Fix create-pr-on-tags to use k tag not v

nimalank7:
Description:
- Currently the create-pr-on-tags action uses the latest 'v' version due to git tag. Changing this to `git tag -l "k*"` will let us use the `k` version
- Future PR will revert this back to `v` once we are satisfied the pipeline is working correctly


kentsanggds (approved):
LGTM

PR #1222, opened 2023-12-05
Add a datagovuk user to the Dockerfile

kentsanggds:
Adding the datagovuk user enables rails db processes

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -19,3 +19,5 @@ RUN gem install bundler && bundle install
 
 RUN chown -R datagovuk:datagovuk $DATAGOVUK_HOME
 USER datagovuk
+
+# TODO: check if we really need the datagovuk user to own DATAGOVUK_HOME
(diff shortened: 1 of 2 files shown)


docker/Dockerfile
```diff
@@ -8,6 +8,14 @@ RUN apt-get install -y nodejs postgresql postgresql-contrib
...
+ENV DATAGOVUK_HOME /srv/app/datagovuk_publish
+
+# Create datagovuk user
+RUN useradd -r -u 900 -m -c "datagovuk account" -d $DATAGOVUK_HOME -s /bin/false datagovuk
+
+COPY ./ $DATAGOVUK_HOME
 
 RUN gem install bundler && bundle install
+
+RUN chown -R datagovuk:datagovuk $DATAGOVUK_HOME
```
sengi, line 20: The app shouldn't need to write to the container image, so it shouldn't need to be the owner of any files in there. Do you know if this is really needed? If it is, then let's call it out as a TODO to fix it later at least.

sengi (approved):

docker/Dockerfile
kentsanggds: I haven't looked into it in too much details but things seem to work from a fresh startup with the new Dockerfile, I'll add a TODO for future fix

PR #1261, opened 2024-02-26
Remove sidekiq schedule and sentry raven

kentsanggds:
The sidekiq schedule is replaced by cronjobs in the govuk-dgu-charts and sentry-raven is handled by the govuk_app_config gem so does not need to be managed in this app.


sengi:
These two things seem unrelated — probably better to make them separate PRs?

sengi:
You could also save a bunch of resources by removing the k8s cronjobs rather than the Sidekiq ones. I don't think the cronjobs being visible buys us much (it's just as easy to see tasks being injected by looking at the worker logs), but it has several small downsides like more components to think about, more config, more different places to look when things go wrong.

kentsanggds:
ok, I can move the cronjob tasks to sidekiq instead, was just following convention so can do it the other way

PR #1267, opened 2024-03-06
Add dev.Dockerfile to use on local dev cluster

kentsanggds:
dev.Dockerfile extends the production Dockerfile to include development environment dependencies that would be useful to use in a local cluster to aid running tests and debugging issues.


docker/dev.Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ghcr.io/alphagov/datagovuk_publish:v2.16.0
+
+USER root
+ENV BUNDLE_WITHOUT=""
+ENV RAILS_ENV=development
+RUN apt update && install_packages \
```
sengi, line 6: [install_packages](https://github.com/alphagov/govuk-ruby-images/blob/main/install_packages.sh#L11) already takes care of the `apt update`.

```suggestion
RUN install_packages \
```

sengi (approved):
Cool 👍

docker/dev.Dockerfile
kentsanggds: ok, I've taken it out

PR #1301, opened 2024-07-19
Update MAX_SIZE deletion to 15647

kentsanggds:
Delete 15647 records as one of the publishers wishes to remove all their datasets.


nimalank7 (approved):
LGTM!

PR #1315, opened 2024-10-03
Fix inspire mapper harvest object

kentsanggds:
The harvest object id from the harvest element in the CKAN response is more up to date than the harvest object id from extras so use that instead.

Also add in a way to build a dev image for debugging on the deployed cluster.


AgaDufrat (approved):
LGTM

